PR #35, opened 2025-01-10
feature: Implement Vulnerability Toggles

Swangeon:
Add GET and POST routes to the toggles page
Add way to see what toggles are on and off through the checkboxes


resources/views/vuln_toggles/index.blade.php
```diff
@@ -0,0 +1,67 @@
...
+        <h2 class="font-semibold text-xl text-gray-800 dark:text-gray-200 leading-tight">
+            {{ __('Vulnerability Toggles') }}
+        </h2>
+    </x-slot>
+
+    <div class="py-12">
+        <div class="max-w-7xl mx-auto sm:px-6 lg:px-8">
+            <div class="bg-white dark:bg-gray-800 overflow-hidden shadow-sm sm:rounded-lg">
+                <div class="p-6 text-gray-900 dark:text-gray-100">
+                    <h1 class="text-2xl font-bold mb-6">Beaver HealthCare</h1>
```
c-a-ray, line 12: I think we'll want to get rid of the extra "Beaver Healthcare" and "Vulnerability Toggles" headings here. It already says "Beaver Healthcare" in the top left corner and "Vulnerability Toggles" is already displayed in the page heading.

resources/views/vuln_toggles/index.blade.php
```diff
@@ -0,0 +1,67 @@
...
+
+                    <h2 class="text-xl font-semibold mb-4">Vulnerability Toggles</h2>
+
+                    <form method="POST" action="{{ route('vulnerability_toggles.store') }}" class="space-y-4">
+                        @csrf
+
+                        <div class="space-y-3">
+                            <div class="flex items-center">
```
c-a-ray, line 20: This could be simplified quite a bit with a `@foreach`. E.g.

```php
 <div class="space-y-3">
    @foreach ([
        'sqli_on' => 'SQL Injection',
        'cmd_inject_on' => 'Command Injection',
        'idor_on' => 'Indirect Object Reference',
        'file_upload_on' => 'File Upload',
        'xss_stored_on' => 'Stored Cross Site Scripting',
        'xss_reflected_on' => 'Reflected Cross Site Scripting'
    ] as $toggle => $label)
        <div class="flex items-center">
            <input
                type="checkbox"
                class="toggle-checkbox"
                id="{{ $toggle }}"
                data-toggle="{{ $toggle }}"
                {{ $user[$toggle] ? 'checked' : '' }}
            >
            <label for="{{ $toggle }}" class="ml-3">{{ $label }}</label>
        </div>
    @endforeach
</div>
```

resources/views/vuln_toggles/index.blade.php
```diff
@@ -0,0 +1,67 @@
...
+                            </div>
+
+                            {{-- <div class="flex items-center">
+                                <input type="checkbox" name="broken_access" id="broken_access" class="w-5 h-5 rounded border-gray-300" {{ $user-> ? 'checked' : '' }}>
+                                <label for="broken_access" class="ml-3">Broken Access Controls</label>
+                            </div> --}}
+                        </div>
+
+                        <div class="mt-6">
+                            <button type="submit" class="px-4 py-2 bg-gray-500 text-white rounded hover:bg-gray-600">
```
c-a-ray, line 57: If you use an AJAX request, you can get rid of the button and it will update automatically when you click a checkbox. That's how I did it in the Admin page. E.g.

```php
    <script>
    document.querySelectorAll('.toggle-checkbox').forEach((checkbox) => {
        checkbox.addEventListener('change', function () {
            const toggle = this.dataset.toggle;
            const isChecked = this.checked;

            fetch('{{ route('vulnerability_toggles.update') }}', {
                method: 'POST',
                headers: {
                    'Content-Type': 'application/json',
                    'X-CSRF-TOKEN': '{{ csrf_token() }}'
                },
                body: JSON.stringify({
                    toggle: toggle,
                    value: isChecked
                })
            })
            .then(response => {
                if (!response.ok) {
                    throw new Error('Response was not ok');
                }
                return response.json();
            })
            .then(data => {
                if (data.success) {
                    console.log('Successfully updated toggle');
                } else {
                    console.error('Failed to update toggle');
                }
            })
            .catch(error => {
                console.error('Fetch error:', error);
            });
        });
    });
    </script>
```

c-a-ray (changes requested):
Nice work, this is looking great. Just a couple things:
- There are duplicated headings for "Beaver Healthcare" and "Vulnerability Toggles". They already appear at the top of the page, in the header section.
- You can get rid of the "Submit" button if you use AJAX requests (like how it is done on the Admin page). If you want to see a working version of how that would work, checkout `cr-vuln-toggle-suggestions`. Up to you, but I think it's an improvement and would make the app more consistent.

c-a-ray (changes requested):
This page is looking really awesome. Super cool, hats off. Good idea moving the JavaScript into resources. I should do the same thing for the JS in the Admin page.

There are just two UX things that make it a little confusing and one deviation from requirements. Hope you don't hate me for this :laughing:. The changes I'm suggesting are very simple, at least.

First, the text above the table says 

> If you want to turn on the secured version of a selected page, check the box next to it and it will be automatically updated.

which makes it sound like checking a checkbox turns the vulnerability off, when it actually turns it on. E.g. when you check the "SQL Injection" checkbox, the database field `user.sqli_on` is set to 1. It would be really confusing for `sqli_on = TRUE` to mean "SQL injection is turned off".

I  really like where you're going with the help text, but I think it would make more sense if it said something like "To enable a vulnerability, check the corresponding checkbox below." Or something like that.

The second thing is the popup. I think it's a great addition, but there are a couple things that make it confusing:
- The messages refer to the toggles by the pages they are associated with rather than the vulnerabilities. E.g., "Successfully updated Change Password toggle." rather than "Successfully updated SQL Injection toggle." Since these are "vulnerability toggles" and not "page toggles" I don't think this makes sense. It's especially problematic since there are two different vulnerabilities for the Patient Feedback page. Both the stored and reflected XSS toggles say "Successfully updated Patient Feedback toggle."
- When a popup appears, it doesn't go away until you refresh the page. This makes it so, for example, if I check a checkbox, then uncheck the same checkbox, there's no indication as to whether the second action did anything. It's also just kind of weird. I think it would make a lot more sense for the popup to disappear after a second or two.
- It might also be beneficial to indicate whether the toggle was turned off or on in the popup. E.g. "SQL Injection is now enabled." and "SQL Injection is now disabled." might be clearer. You could get even more explicit, too, and say something like, "SQL Injection is now enabled in the Change Password page.", etc.

Last issue is with the IDOR toggle. Three things:
- It's "Insecure Direct Object Reference," not "Indirect Object Reference."
- We decided to change the Employee Information page to the Patient Information page.
- The IDOR toggle is for both the Admin page and the Patient Information page.

Let me know if you want to get together to chat about any of this stuff.

c-a-ray (approved):
Nice!

PR #36, opened 2025-01-12
Some minor refactoring for clarity

c-a-ray:
This PR refactors a few things to promote clarity and consistency. Main things:
- More descriptive names for the `AuthenticatedSessionController`
- Use `faker` in the `UserFactory`
- Remove the "Remember Me" checkbox from the login form (leftover from Breeze default, not in our original design)
- Simplify `postbuild.js`
- Clean up / consolidate the routes in `web.php`
- Update the way we mount project directories in `docker-compose.yaml` (to fix an issue with the `build` directory getting overwritten)

Note that there will be conflicts in `web.php` with #28 and #35. Once those PRs get merged, I'll update this to fix the conflicts.


bootstrap/app.php
```diff
@@ -7,12 +7,15 @@
...
-        commands: __DIR__.'/../routes/console.php',
         health: '/up',
     )
-    ->withMiddleware(function (Middleware $middleware) {
-        //
-    })
-    ->withExceptions(function (Exceptions $exceptions) {
-        //
-    })->create();
+    ->withMiddleware(
```
c-a-ray, line 12: I tried removing these (because we don't need any global middleware at this point), but apparently they need to be there. The changes are just from the formatter.

bootstrap/app.php
```diff
@@ -7,12 +7,15 @@
 return Application::configure(basePath: dirname(__DIR__))
     ->withRouting(
         web: __DIR__.'/../routes/web.php',
-        commands: __DIR__.'/../routes/console.php',
```
c-a-ray, line 10: This file was removed a while ago because we didn't have a use for it.

config/auth.php
```diff
@@ -14,8 +14,8 @@
     */
 
     'defaults' => [
-        'guard' => env('AUTH_GUARD', 'web'),
```
c-a-ray, line 17: Removed these while debugging something and left it this way because it seems easier to read. If we end up using environment variables we can add them back in.

database/factories/UserFactory.php
```diff
@@ -7,18 +7,16 @@
...
 class UserFactory extends Factory
 {
-    protected static ?string $password = null;
-
     public function definition(): array
     {
         return [
-            'username' => fake()->unique()->userName(),
-            'password' => static::$password ??= Hash::make('password'),
+            'username' => $this->faker->unique()->userName,
```
c-a-ray, line 13: I suggested using `faker` in the Patient Info page PR, then realized we weren't using it in the `UserFactory`. Changed for consistency. Also allows us to assign different roles to different users. Here, each role has a 20% chance of being true.

resources/views/auth/login.blade.php
```diff
@@ -24,14 +24,6 @@
             <x-input-error :messages="$errors->get('password')" class="mt-2" />
         </div>
 
-        <!-- Remember Me -->
```
c-a-ray, line 27: This wasn't part of our design and IMO complicates things unnecessarily.

routes/web.php
```diff
@@ -8,58 +8,49 @@
...
 use App\Http\Controllers\Auth\AuthenticatedSessionController;
+use App\Http\Controllers\Auth\RegisteredUserController;
+use App\Http\Controllers\Auth\NewPasswordController;
+use App\Http\Controllers\Auth\PasswordController;
 use Illuminate\Support\Facades\Route;
 
 
-// Login
-Route::get('/', [AuthenticatedSessionController::class, 'create']);
+Route::middleware('guest')->group(
```
c-a-ray, line 17: Seems easier to keep track of everything by keeping all of the routes in one file, so I moved the routes from `auth.php` here.

routes/web.php
```diff
@@ -8,58 +8,49 @@
...
-Route::get('/dashboard', [DashboardController::class, 'index'])
-    ->middleware(['auth', 'verified'])
-    ->name('dashboard');
+        Route::get('reset-password/{token}', [NewPasswordController::class, 'create'])->name('password.reset');
+        Route::post('reset-password', [NewPasswordController::class, 'store'])->name('password.store');
+    }
+);
 
-// Profile
 Route::middleware('auth')->group(
```
c-a-ray, line 31: The `verified` middleware is to make sure the user has had their e-mail verified. Since we're using `username` instead of `email`, we don't need it.

Since all of these routes just require the `auth` middleware, they can be grouped together, which IMO makes this file a ton easier to read.

scripts/postbuild.js
```diff
@@ -2,56 +2,53 @@ import { promises as fs } from 'fs';
 import path from 'path';
```
c-a-ray, line 2: The sloppiness of this script was bugging me. It does the exact same thing now, but is a lot easier to follow.

docker-compose.yaml
```diff
@@ -18,7 +18,12 @@ services:
     depends_on:
       - db
     volumes:
-      - .:/var/www/html:delegated
+      - ./app:/var/www/html/app:delegated
```
c-a-ray, line 21: I was having issues with the build that turned out to be a result of us mounting the entire project directory in the container. Specifically, if `public/build` didn't exist on the host machine, it wouldn't be created in the container. To fix it I had to go into the container and manually run `npm install && npm run build`.

The reason we need to mount any of this stuff at all is only for development. It allows us to not have to rebuild the container any time we make a change. By only mounting the directories needed to make that happen, the `build` directory no longer gets overwritten.

database/factories/UserFactory.php
alherrera-cs: Nice work updating the UserFactory to use Faker! Since I used this on the PatientFactory with $this->faker for generating realistic data, it's great to see that the UserFactory now follows a similar approach. Do you think our app requires that we link users with patients at all, or will they continue to remain separate entities?

docker-compose.yaml
alherrera-cs: Great improvement on the docker-compose.yaml! Mounting only the necessary directories instead of the entire project fixes the build directory issues, which have caused me a lot of headaches in development. I appreciate the extra clarification.

alherrera-cs:
Great work—thank you for always paying close attention to detail and ensuring our app leaves out unnecessary or confusing components. Your contributions consistently help streamline our codebase, and I really appreciate the clarity you bring to these changes. Thank you!

alherrera-cs (commented):
Great work—thank you for always paying close attention to detail and ensuring our app leaves out unnecessary or confusing components. Your contributions consistently help streamline our codebase, and I really appreciate the clarity you bring to these changes. Thank you!

resources/views/components/sidebar.blade.php
```diff
@@ -28,15 +28,15 @@
             @endif
 
             <!-- Vulnerability Toggles Link -->
-            <a href="{{ route('vulnerability_toggles') }}" class="block py-2 px-4 rounded hover:bg-gray-600">Vulnerability Toggles</a>
+            <a href="{{ route('toggles') }}" class="block py-2 px-4 rounded hover:bg-gray-600">Vulnerability Toggles</a>
 
             <!-- Profile Link -->
-            <a href="{{ route('profile.edit') }}" class="block py-2 px-4 bg-gray-700 rounded hover:bg-gray-600">Profile</a>
+            <a href="{{ route('profile.edit') }}" class="block py-2 px-4 rounded hover:bg-gray-600">Profile</a>
```
c-a-ray, line 34: Currently the Profile and Log Out links are always highlighted. This fixes that.

scripts/postbuild.js
```diff
@@ -2,56 +2,53 @@ import { promises as fs } from 'fs';
...
+async function fileExists(filePath) {
+  try {
+    await fs.stat(filePath);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+ensureManifestLocation();
```
Swangeon, line 54: I believe `fs.existsSync()` does this already so we can switch to that unless we can't use it here for some reason.

Swangeon (changes requested):
Solid work! Just wanted to see if you could use `fs.existsSync()` instead of making a whole function for finding a file in [postbuild.js](https://github.com/BeaverHealth-Vulnerable-Web-App/BeaverHealth-Vulnerable-Web-App/pull/36/files#diff-1e428174cfde260659e43151349d38eb8030ebd486f36cfdcab52438ef40e09c).

scripts/postbuild.js
c-a-ray: Nice, I didn't know about this method. Updated the function to use it. Also added in some logs and errors to make it easier to debug if something goes wrong.

PR #49, opened 2025-01-19
File Upload Page

BrynnMartina:
Created view and controller
Added POST in web.php
Uploaded files get stored in storage > app > public > uploads

### DIFF
--- a/app/Http/Controllers/AddRecordsController.php
+++ b/app/Http/Controllers/AddRecordsController.php
@@ -2,10 +2,33 @@
 
 namespace App\Http\Controllers;
 
+use Illuminate\Http\Request;
+
 class AddRecordsController extends Controller
 {
     public function index()
     {
-        return view('dashboard'); // Update this with the actual view for add records page
+        return view('records.add-records'); // Add records page in views > records
+    }
+
+    // File upload
+    public function upload(Request $request)
+    {   
+        // Allow only files (no other validation for now)
+        $request->validate([
+            'file' => 'required|file',
+        ]);
+
+        // Retrieve the uploaded file
+        $file = $request->file('file');
+
+        // Get filename
+        $filename = $file->getClientOriginalName();
+
+        // Store the file in storage/app/public/uploads
+        $path = $file->storeAs('uploads', $filename, 'public');
+
+        return back()->with('success', 'File uploaded successfully!');
     }
 }
+
--- a/app/Http/Controllers/AddRecordsController.php
+++ b/app/Http/Controllers/AddRecordsController.php
@@ -8,7 +8,7 @@ class AddRecordsController extends Controller
 {
     public function index()
     {
-        return view('records.add-records'); // Add records page in views > records
+        return view('records.add-records');
     }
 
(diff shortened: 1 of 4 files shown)


resources/views/records/add-records.blade.php
```diff
@@ -0,0 +1,33 @@
...
+
+    <div class="py-12">
+        <div class="max-w-7xl mx-auto sm:px-6 lg:px-8">
+            <div class="bg-white dark:bg-gray-800 overflow-hidden shadow-sm sm:rounded-lg">
+                <div class="p-6 text-gray-900 dark:text-gray-100">
+                    <!-- File upload form -->
+                    <form action="{{ route('upload') }}" method="POST" enctype="multipart/form-data">
+                        @csrf
+                        <div class="mb-4">
+                            <label for="file" class="block mb-2 font-semibold text-gray-800 dark:text-gray-200">Choose a file (PDFs Only):</label>
```
c-a-ray, line 16: I would get rid of the "(PDFs Only)" part of this, since you can upload text files, etc., as well

app/Http/Controllers/AddRecordsController.php
```diff
@@ -2,10 +2,33 @@
...
 namespace App\Http\Controllers;
 
+use Illuminate\Http\Request;
+
 class AddRecordsController extends Controller
 {
     public function index()
     {
-        return view('dashboard'); // Update this with the actual view for add records page
+        return view('records.add-records'); // Add records page in views > records
```
c-a-ray, line 11: Some of these comments are a little unnecessary IMO. I'd get rid of the comments on lines 11, 14, and 22, just because the code is pretty self-explanatory. The comments on lines 17 and 28 are helpful though

c-a-ray (commented):
Nice work, just two small things:
- I don't think it should say "PDFs Only" on the file upload page
- Some of the comments in AddRecordsController are a little redundant

app/Http/Controllers/AddRecordsController.php
BrynnMartina, line 1: Removed `return view('dashboard');` and comments

resources/views/records/add-records.blade.php
```diff
@@ -0,0 +1,33 @@
...
+
+    <div class="py-12">
+        <div class="max-w-7xl mx-auto sm:px-6 lg:px-8">
+            <div class="bg-white dark:bg-gray-800 overflow-hidden shadow-sm sm:rounded-lg">
+                <div class="p-6 text-gray-900 dark:text-gray-100">
+                    <!-- File upload form -->
+                    <form action="{{ route('upload') }}" method="POST" enctype="multipart/form-data">
+                        @csrf
+                        <div class="mb-4">
+                            <label for="file" class="block mb-2 font-semibold text-gray-800 dark:text-gray-200">Choose a file (PDF, TXT, or CSV):</label>
```
BrynnMartina, line 16: Changed from "PDF Only" to "PDF, TXT, or CSV"

c-a-ray (approved):
Nice work

PR #54, opened 2025-01-24
testing: Creating vulnerability toggles feature tests

Swangeon:
- Run with `./vendor/bin/sail test`


Swangeon:
# Notes
- The testing methods have to be in snake case because of how they appear when running pest.
- The `auth` middleware makes testing the controllers a little difficult so the user attribute in the VulnTogglesController had to become a public attribute of the class so that the functions would work.

tests/Unit/VulnTogglesTest.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+use App\Http\Controllers\VulnerabilityTogglesController;
+use App\Models\User;
+use Illuminate\Testing\TestView;
+use Tests\TestCase;
+
+
+class VulnTogglesUnitTest extends TestCase
```
c-a-ray, line 9: These tests aren't truly unit tests because they rely on the database by using `User::factory()->create()` and rely on the authentication system by using `$this->actingAs($this->user)`. Also, manually setting `$this->controller->user = $this->user` bypasses the controller's logic for retrieving the user.

To properly isolate things and make the controller testable, you'll need to update the controller (as I suggested above) to retrieve the user from the `Request`. This will allow you to inject a mocked `User` instance directly through the `Request` object, which removes the dependencies on the database and the authentication system.

Below is just an example. I didn't have enough time to go through this as thoroughly as I did the feature tests.
```php
<?php

namespace Tests\Unit;

use Tests\TestCase;
use App\Http\Controllers\VulnerabilityTogglesController;
use App\Models\User;
use Illuminate\Http\Request;
use Illuminate\View\View;
use Mockery;

class VulnTogglesUnitTest extends TestCase
{
    protected $controller;
    protected $userMock;
    protected $requestMock;

    protected function setUp(): void
    {
        parent::setUp();

        $this->controller = new VulnerabilityTogglesController();
        $this->userMock = Mockery::mock(User::class);
        $this->requestMock = Mockery::mock(Request::class);
    }

    protected function tearDown(): void
    {
        Mockery::close();
        parent::tearDown();
    }

    public function test_vuln_toggles_index_returns_correct_view()
    {
        $this->requestMock
            ->shouldReceive('user')
            ->once()
            ->andReturn($this->userMock);

        $response = $this->controller->index($this->requestMock);

        $this->assertInstanceOf(View::class, $response, 'Response is not an instance of View');
        $this->assertEquals('vuln_toggles.index', $response->name(), 'View name does not match');

        $viewData = $response->getData();
        $this->assertArrayHasKey('user', $viewData, 'View data does not have "user" key');
        $this->assertSame($this->userMock, $viewData['user'], 'User data does not match the mocked user');
    }
}
```

Also, the tests should cover a wider range of scenarios (including edge cases and validation failures).

tests/Feature/FeedbackTest.php
```diff
@@ -1,51 +1,62 @@
...
 
-function sendTestFeedback($user, $testObject) {
-    $feedbackData = [
-        'fname' => 'Test',
-        'lname' => 'User',
-        'feedback' => 'This is a test comment.'
-    ];
-    return $testObject->actingAs($user)->post('/feedback/store', $feedbackData);
-}
+class FeedbackFeatureTest extends TestCase
```
c-a-ray, line 7: There are some issues here that are similar to what is pointed out in the vuln toggle feature test file, but since these were pre-existing, I'd be cool with leaving it for now and updating it in the future.

c-a-ray (changes requested):
Nice job on all this! Going through this PR was a huge learning experience for me. I've learned a ton about Laravel today. Hopefully my comments aren't overwhelming. I tried to give working examples as much as I could to make things easier on you. Let me know if there's anything I can help with. 

Totally up to you, but we might consider just focusing on the feature tests for this PR and push the unit tests to next week. There's just a lot to consider here.

c-a-ray:
I noticed in your PR description that it says "Run with `./vendor/bin/sail pest`". While this works for some reason, since we aren't using Pest I think it makes more sense to use `./vendor/bin/sail test`

Swangeon:
> I noticed in your PR description that it says "Run with `./vendor/bin/sail pest`". While this works for some reason, since we aren't using Pest I think it makes more sense to use `./vendor/bin/sail test`

Pest runs on top PHPUnit which is why it can run them and using Pest allows for keeping a config within [Pest.php](https://github.com/BeaverHealth-Vulnerable-Web-App/BeaverHealth-Vulnerable-Web-App/blob/062babc5fcad3abbce8c08fb6225dbe1c889a389/tests/Pest.php#L14) that allows all tests to be run in a transaction.

alherrera-cs (approved):
Great job implementing the VulnTogglesService to handle business logic separately from the controller, following Cody's suggestions from our meeting yesterday. This makes the controller cleaner and easier to maintain, and simplifies the unit testing process. Your thorough feature tests for vulnerability toggles significantly boost our application's reliability and security. Great job Sean!

tests/Unit/ExampleTest.php
```diff
@@ -1,5 +0,0 @@
-<?php
```
c-a-ray, line 1: We should leave this file here as a placeholder until we have real unit tests. Since this is the only file in `tests/Unit/`, deleting it removes the entire directory from Git, which causes `./vendor/bin/sail test` to fail because PHPUnit expects the `tests/Unit/` directory to exist.

c-a-ray, line 1: We should leave this file here as a placeholder until we have real unit tests. Since this is the only file in `tests/Unit/`, deleting it removes the entire directory from Git, which causes `./vendor/bin/sail test` to fail because PHPUnit expects the `tests/Unit/` directory to exist.

tests/Feature/VulnTogglesFeatureTest.php
```diff
@@ -0,0 +1,163 @@
...
+            ['xss_reflected_on'],
+        ];
+    }
+
+    public function sendValidTestRequest($toggle, $value)
+    {
+        $checkData = [
+            'toggle' => $toggle,
+            'value' => $value,
+            '_token' => csrf_token(),
```
c-a-ray, line 42: In PR #57 I added a `postWithCsrf` method to the `TestCase` class so we don't have to manually include the token for post requests like this.

```php
    protected function postWithCsrf(string $route, array $data = []): \Illuminate\Testing\TestResponse
    {
        return $this->post($route, array_merge(['_token' => session()->token()], $data));
    }
```

If you copy this over into your branch, you can just do:
```php
$this->postWithCsrf(
    route('vulnerability_toggles.update'), [
        'toggle' => $toggle,
        'value' => $value,
    ]
);
```

tests/Feature/VulnTogglesFeatureTest.php
```diff
@@ -0,0 +1,163 @@
...
+        ];
+        return $this->post('/vulnerability_toggles/update', $checkData);
```
c-a-ray, line 44: Consider using the named route (`route('vulnerability_toggles.update)`) instead of the hardcoded URL. Avoiding direct URL references improves maintainability.

tests/Feature/VulnTogglesFeatureTest.php
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    public function test_vuln_toggles_route()
+    {
+        $response = $this->get('/vulnerability_toggles');
+        $response->assertStatus(self::HTTP_OK);
```
c-a-ray, line 50: You can just use `$response->assertOk()`

tests/Feature/VulnTogglesFeatureTest.php
```diff
@@ -0,0 +1,163 @@
...
+            ['sqli_on'],
+            ['cmd_inject_on'],
+            ['idor_on'],
+            ['file_upload_on'],
+            ['xss_stored_on'],
+            ['xss_reflected_on'],
+        ];
+    }
+
+    public function sendValidTestRequest($toggle, $value)
```
c-a-ray, line 37: Couple issues here.

It should really be named something like `sendUpdateToggleRequest()`. Calling it `sendValidTestRequest()` is vague and misleading. It is used to send invalid requests in `test_invalid_toggle_name_request()`, `test_invalid_toggle_type_request()`, and `test_invalid_value_type_request()`, so `sendValidTestRequest()` is not really accurate.

Should also be `private` rather than `public` since it's only used inside of this class. There's no reason for it to be publicly accessible.

If you also use the `postWithCsrf` method I talk about in another [comment](https://github.com/BeaverHealth-Vulnerable-Web-App/BeaverHealth-Vulnerable-Web-App/pull/54/files#r1934756082), this can become a clean one-liner:
```php
    private function sendToggleUpdateRequest($toggle, $value)
    {
        return $this->postWithCsrf(route('vulnerability_toggles.update'), compact('toggle', 'value'));
    }
```

tests/Feature/VulnTogglesFeatureTest.php
```diff
@@ -0,0 +1,163 @@
...
+        $response->assertStatus(self::HTTP_OK);
+    }
+
+    #[DataProvider('toggleProvider')]
+    public function test_valid_toggle_on($toggle)
+    {
+        $response = $this->sendValidTestRequest($toggle, true);
+        $response->assertStatus(self::HTTP_OK);
+
+        $response = $this->get('/vulnerability_toggles');
```
c-a-ray, line 59: Doesn't really make sense to do a `GET` request _after_ the `POST` request. The purpose of a feature/integration test is to simulate real user interactions as they would happen in production. In a real scenario, a user would have to first make a `GET`  request to load the vuln toggles page before updating a toggle. 

An added benefit of this is that the `GET` request starts the session, so you wouldn't need the `session()->start()` in the `setUp()` method. Really should try to avoid that as much as possible since it bypasses Laravel's normal request lifecycle.

tests/Feature/VulnTogglesFeatureTest.php
```diff
@@ -0,0 +1,163 @@
...
+    {
+        $response = $this->sendValidTestRequest('invalid_toggle', true);
+        $response->assertStatus(self::HTTP_OK);
+        $response->assertJson(['success' => false]);
+    }
+
+    public function test_invalid_toggle_type_request()
+    {
+        $response = $this->sendValidTestRequest(1234, true);
+        $response->assertStatus(self::HTTP_UNPROCESSABLE_ENTITY);
```
c-a-ray, line 153: Can just use `$response->assertUnprocessable()`

tests/Feature/VulnTogglesFeatureTest.php
```diff
@@ -0,0 +1,163 @@
...
+{
+    protected $user;
+    protected const HTTP_OK = 200;
+    protected const CSRF_TOKEN_MISMATCH = 419;
+    protected const HTTP_UNPROCESSABLE_ENTITY = 422;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->user = User::factory()->create();
```
c-a-ray, line 20: Storing `$user` as an attribute and handling authentication in `setUp()` isn't ideal. Not every test requires an authenticated user. It also introduces shared state across tests, which makes them less isolated and potentially flaky. IMO user creation should be done inside each test that actually needs authentication. Would make the test requirements more explicit and improve readability.

tests/Feature/VulnTogglesFeatureTest.php
```diff
@@ -0,0 +1,163 @@
...
+        $this->actingAs($this->user);
+        session()->start();
```
c-a-ray, line 22: I talk about this in another comment somewhere, but I don't think having `session()->start()` in `setUp()` is ideal.

Most of the tests here have a `GET` requests in them, and the `GET` request starts a session, so you can just move the `GET` requests before the `POST` requests (which is what would happen in a real user interaction anyway).

Also, I learned that we shouldn't be using `session()->start()` in tests at all, but rather `$this->startSession()`, which is a test helper function that properly initializes the session within the test lifecycle. 

IMO the only test here that seems like it might make sense to start a session like this is in `test_unauthenticated_vuln_toggles_update()` since the user is logged out before making the `POST` request.

tests/Feature/VulnTogglesFeatureTest.php
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    public function test_non_csrf_request()
+    {
+        $checkData = [
+            'toggle' => 'sqli_on',
+            'value' => true
+        ];
+        $response = $this->post('/vulnerability_toggles/update', $checkData);
+        $response->assertStatus(self::CSRF_TOKEN_MISMATCH);
```
c-a-ray, line 140: There isn't an `assertCsrfToken()` method in `AssertsStatusCodes` for some reason, but I created it in PR #57. Check out the `TestCase.php` on that branch.

tests/Feature/VulnTogglesFeatureTest.php
```diff
@@ -0,0 +1,163 @@
...
+                'user_id' => $this->user->user_id,
+                $toggle => false
+            ]
+        );
+
+        $response = $this->get('/vulnerability_toggles');
+        $response->assertStatus(self::HTTP_OK);
+
+        $crawler = new Crawler($response->content());
+        $checkbox = $crawler->filter("input[type='checkbox'][id='" . $toggle . "']")->first();
```
c-a-ray, line 94: Nice job figuring a better query for the crawler filter. Only thing is I don't think the `->first()` should be there.

tests/Feature/FeedbackFeatureTest.php
```diff
@@ -0,0 +1,64 @@
...
+
+    public function test_invalid_feedback_search()
+    {
+        $response = $this->sendTestFeedback();
+        $response->assertStatus(302);
+        $response = $this->get('/feedback/search?search_name=admin');
+        $response->assertStatus(200);
+        $response->assertDontSeeText('This is a test comment.');
+    }
+}
```
c-a-ray, line 64: A lot of the things I mentioned in `VulnTogglesFeatureTest.php` apply here too. Won't belabor them here.

app/Http/Requests/VulnTogglesRequest.php
```diff
@@ -0,0 +1,44 @@
...
+namespace App\Http\Requests;
+
+use Illuminate\Contracts\Validation\Validator;
+use Illuminate\Foundation\Http\FormRequest;
+use Illuminate\Http\Exceptions\HttpResponseException;
+use Illuminate\Http\JsonResponse;
+
+class VulnTogglesRequest extends FormRequest
+{
+    /**
```
c-a-ray, line 12: Not saying remove these, but we should have a team conversation about PHPDoc blocks and decide if we want to use them. I just think we should be consistent.

app/Http/Controllers/VulnerabilityTogglesController.php
```diff
@@ -2,36 +2,18 @@
```
c-a-ray, line 2: This is awesome. You've got the controller purely focusing on handling HTTP requests.

app/Http/Controllers/VulnerabilityTogglesController.php
```diff
@@ -2,36 +2,18 @@
...
-            'value' => 'required|boolean',
-            ]
-        );
-
-        $toggle = $request->input('toggle');
-        $value = $request->input('value') ? true : false;
-        $this->user->{$toggle} = $value;
-        $this->user->save();
-        return response()->json(['success' => true]);
+        return (new VulnTogglesService())->updateVulnToggles($request);
```
c-a-ray, line 17: The only improvement I would make to this file is to not manually instantiate the service inside of the method. It just prevents dependency injection and will make writing unit tests harder.

From what I've seen, this is how you're supposed to do this:

```php
class VulnerabilityTogglesController extends Controller
{
    private VulnTogglesService $vulnTogglesService;

    public function __construct(VulnTogglesService $vulnTogglesService)
    {
        $this->vulnTogglesService = $vulnTogglesService;
    }

    public function index(): View
    {
        return view('vuln_toggles.index', ['user' => auth()->user()]);
    }

    public function update(VulnTogglesRequest $request): JsonResponse
    {
        return $this->vulnTogglesService->updateVulnToggles($request);
    }
}
```

app/Services/VulnTogglesService.php
```diff
@@ -0,0 +1,21 @@
+<?php
```
c-a-ray, line 1: Nice, our first service class!

app/Services/VulnTogglesService.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace App\Services;
+
+use App\Http\Requests\VulnTogglesRequest;
+
+class VulnTogglesService
+{
+    public function updateVulnToggles(VulnTogglesRequest $request)
+    {
+        $toggle = $request->input('toggle');
```
c-a-ray, line 11: Since you're making multiple calls to `auth()->user()`, might as well store it in a variable:
```php
$user = auth()->user();
```

app/Services/VulnTogglesService.php
```diff
@@ -0,0 +1,21 @@
...
+        $value = $request->input('value') ? true : false;
+        if (array_key_exists($toggle, auth()->user()->getAttributes())) {
+            auth()->user()->{$toggle} = $value;
```
c-a-ray, line 14: I just learned about this, but you can combine these two lines and make this a bit more readable with
```php
$user->update([$toggle => $value]);
```

app/Services/VulnTogglesService.php
```diff
@@ -0,0 +1,21 @@
...
+            auth()->user()->save();
+            return response()->json(['success' => true]);
+        } else {
+            return response()->json(['success' => false]);
```
c-a-ray, line 18: Consider including a useful error message on failure. E.g.,
```suggestion
            return response()->json(['success' => false, 'error' => 'Invalid toggle name'], 422);
```

app/Services/VulnTogglesService.php
c-a-ray, line 14: I just learned about this, but you can combine these two lines and make this a bit more readable with
```php
$user->update([$toggle => $value]);
```

c-a-ray (changes requested):
This is really coming together! Really, nice work. The controller is looking much cleaner and the tests have a lot more coverage.

Probably seems like I left a ton of comments, but it's all really small stuff. Also, you're obviously not obligated to agree with everything or make every single change I suggest. My only goal with these reviews is to produce the best possible codebase, and it's pretty much guaranteed that I haven't thought of everything, so please push back if you see a better way to approach something.

app/Http/Controllers/VulnerabilityTogglesController.php
Swangeon: I really gotta get a new PHP LSP because I tried that before and it kept saying that it was an error to do so 🤦. Will change it though!

tests/Feature/VulnTogglesFeatureTest.php
Swangeon: I agree with you on your second point but the reason why that *GET* request is there after the *POST* is for checking to make sure that the toggle is visually updated on the page as a *GET* request before the *POST* request wouldn't update the View that is returned from that initial *GET*. At the very least, it hasn't updated in my testing.

tests/Feature/VulnTogglesFeatureTest.php
Swangeon: I personally think it should still be here as the amount of tests that need a authenticated state is 8/10 and I feel would make the code less DRY compliant. I do think that being authenticated is the default state for this test and the shared user state isn't causing problems.

tests/Feature/FeedbackFeatureTest.php
```diff
@@ -0,0 +1,73 @@
...
+
+    public function test_valid_feedback_search(): void
+    {
+        $response = $this->get(route('feedback'));
+        $response->assertOk();
+
+        $response = $this->sendTestFeedback();
+        $response->assertRedirect(route('feedback'));
+
+        $response = $this->get('/feedback/search?search_name=Test');
```
c-a-ray, line 56: Continuation of the above comment. Just wanted to let you know you can also stack the asserts like this:
```php
        $this->get('/feedback/search?search_name=Test')
            ->assertOk()
            ->assertSeeText('This is a test comment.');
```

tests/Unit/ExampleTest.php
```diff
@@ -1,5 +1,13 @@
 <?php
 
-test('that true is true', function () {
-    expect(true)->toBeTrue();
-});
+namespace Tests\Unit;
```
c-a-ray, line 3: Nice

c-a-ray (changes requested):
Nice work.

The only thing that _needs_ to happen before this gets merged is to fix CI. 

I made a handful of comments about ways you could improve readability, but I wouldn't hold up this PR for that.

.github/workflows/ci.yaml
```diff
@@ -35,3 +31,6 @@ jobs:
 
       - name: Run Laravel Application
         run: vendor/bin/sail up -d --build
+
+      - name: Run Tests
```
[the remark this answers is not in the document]
Swangeon: agreed

README.md
```diff
@@ -1,12 +1,16 @@
 # BeaverHealth Vulnerable Web App
+
 ## Running the Application
+
 1. Install dependencies
+
 - [PHP](https://www.php.net/manual/en/install.php)
+- XML & Curl PHP Modules through your package manager.
```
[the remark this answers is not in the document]
Swangeon: When I was doing the reinstall of the application on my new development machine, I needed to install those to just run the application, not for the tests. I will also try to change the wording.

tests/Feature/VulnTogglesFeatureTest.php
c-a-ray: Sounds good

tests/Feature/VulnTogglesFeatureTest.php
```diff
@@ -0,0 +1,168 @@
...
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->user = User::factory()->create();
+        $this->actingAs($this->user);
+    }
+
+    static function toggleProvider(): array
+    {
+        return [
```
c-a-ray, line 24: I just realized that the the data provider makes it so the tests have output like this:
```sh
  ✓ valid toggle on with data set #0                                                                                                                                                                                                  0.01s
  ✓ valid toggle on with data set #1                                                                                                                                                                                                  0.01s
  ✓ valid toggle on with data set #2                                                                                                                                                                                                  0.01s
  ✓ valid toggle on with data set #3                                                                                                                                                                                                  0.01s
  ✓ valid toggle on with data set #4                                                                                                                                                                                                  0.01s
  ✓ valid toggle on with data set #5                                                                                                                                                                                                  0.01s
  ✓ valid toggle off with data set #0                                                                                                                                                                                                 0.01s
  ✓ valid toggle off with data set #1                                                                                                                                                                                                 0.01s
  ✓ valid toggle off with data set #2                                                                                                                                                                                                 0.01s
  ✓ valid toggle off with data set #3                                                                                                                                                                                                 0.01s
  ✓ valid toggle off with data set #4                                                                                                                                                                                                 0.01s
  ✓ valid toggle off with data set #5                                                                                                                                                                                                 0.01s
```

I just found that if you change it to this
```php
    static function toggleProvider(): array
    {
        return [
            'sqli_on' => ['sqli_on'],
            'cmd_inject_on' => ['cmd_inject_on'],
            'idor_on' => ['idor_on'],
            'file_upload_on' => ['file_upload_on'],
            'xss_stored_on' => ['xss_stored_on'],
            'xss_reflected_on' => ['xss_reflected_on'],
        ];
    }
```
you get this instead

```sh
  ✓ valid toggle on with data set "sqli on"                                                                                                                                                                                           0.01s
  ✓ valid toggle on with data set "cmd inject on"                                                                                                                                                                                     0.01s
  ✓ valid toggle on with data set "idor on"                                                                                                                                                                                           0.01s
  ✓ valid toggle on with data set "file upload on"                                                                                                                                                                                    0.01s
  ✓ valid toggle on with data set "xss stored on"                                                                                                                                                                                     0.01s
  ✓ valid toggle on with data set "xss reflected on"                                                                                                                                                                                  0.01s
  ✓ valid toggle off with data set "sqli on"                                                                                                                                                                                          0.01s
  ✓ valid toggle off with data set "cmd inject on"                                                                                                                                                                                    0.01s
  ✓ valid toggle off with data set "idor on"                                                                                                                                                                                          0.01s
  ✓ valid toggle off with data set "file upload on"                                                                                                                                                                                   0.01s
  ✓ valid toggle off with data set "xss stored on"                                                                                                                                                                                    0.01s
  ✓ valid toggle off with data set "xss reflected on"                                                                                                                                                                                 0.01s
```

c-a-ray (commented):
Good stuff, this is looking great.

Found one little thing. Up to you whether to implement it. Either way, I think this is ready to get merged in.

c-a-ray (approved):
Nice work on this. You're awesome!

PR #55, opened 2025-01-26
Request Records Page

BrynnMartina:
Features (subject to change):
-Input first name, last name, and DOB.
-If the name and DOB match someone in the database, the page will return with the patient's data, as well as any files associated with the patient.
-Currently, associated patient files are stored to storage/app/public/uploads. You have to manually add the files (via accessing the folder, or utilizing the add records page) as they currently aren't pre-populated.
-Associated files have to be formatted as "Firstname Lastname.fileextension"
-For example: "Ray Botsford.pdf"

Vulnerability Features:
-Command injection works in the `First Name:` box. The vulnerability will pop up in the `Medical Files:` output.
-For example, in `First Name:` insert `John; whoami` or `John; ls`

Not implemented:
-Associated files will have their names displayed in `Medical Files:` but are currently not clickable hyperlinks.

### DIFF
--- a/app/Http/Controllers/RequestRecordsController.php
+++ b/app/Http/Controllers/RequestRecordsController.php
@@ -2,10 +2,77 @@
 
 namespace App\Http\Controllers;
 
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\DB;
+
 class RequestRecordsController extends Controller
 {
     public function index()
     {
-        return view('dashboard'); // Update this with the actual view for record requests page
+        // Default "N/A" values for patient data
+        $results = session('results', collect([
+            (object) [
+                'first_name' => 'N/A',
+                'last_name' => 'N/A',
+                'date_of_birth' => 'N/A',
+                'policy_number' => 'N/A',
+                'address' => 'N/A',
+            ]
+        ]));
+
+        // Default value for log output
+        $logOutput = session('log_output', 'N/A');
+
+        return view('records.request', compact('results', 'logOutput'));
+    }
+
+    public function search(Request $request)
+    {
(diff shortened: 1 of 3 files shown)


c-a-ray (changes requested):
Nice job on this! Really, the page does look really good and things are generally pretty well-structured.

I know code freeze is tonight, so I feel bad about not being able to just approve this, but there are definitely some things it would be good to sort out. I wish we could have started the review process earlier so we could have worked this stuff out in time for v0.4.

The biggest thing is that this strays pretty far from the design. If you go back and look at the Design Document, you'll see that this page was supposed to focus on downloading/interacting with the files ("medical records") that were uploaded in the "Add Records" page. As-is, there's no mechanism for doing that. It only lets you search for filenames that contain the patient's first and last name. Also, including patient information wasn't part of the original scope and overlaps with the purpose of the "Patient Information" page. 

From the Design Document:

> The Medical Records Request page will consist of a text input field labeled “Name of Patient” and a submit button. When a user enters the name of a patient and clicks the submit button, the files that were uploaded for that patient in the Add Medical Records page will be downloaded by the client.

Here was the wireframe:
image

I know that was just a rough, preliminary design, so I'm definitely not saying we have to stick to it exactly, but this does deviate pretty significantly from the original scope.

We should talk more about the design in our meeting on Tuesday and get synced up on all this. I have some ideas for how we could make this make more sense. One idea I had while going through this is that we could store files in directories that have the `patient_id` as the directory name. That way the files don't have to have the patient's first and last name as the filename. We'd need to add some text inputs to the "Add Records" page to do that, but that actually was part of the original design too.

PR #58, opened 2025-02-02
Enforce PSR-12 Standards

c-a-ray:
This PR:
- Adds a PHP Code Sniffer (`phpcs`) configuration file (`phpcs.xml`) to enforce PSR-12 coding standards
- Applies PHP Code Beautifier and Fixer (`phpcbf`) to fix code style issues in relevant directories

To ensure consistency moving forward:
- Run `phpcbf` to automatically fix formatting issues before committing code
- Use `phpcs` to check for code style violations that couldn't be fixed automatically by `phpcbf`

Note that we were previously operating under the assumption that PHPUnit tests needed to be in snake case. This turns out to not be true, and actually violates PSR-12, so all test method names have been updated to camel case.

PR also adds a `CONTRIBUTING.md` file with information about development environment setup, including installing PHP_CodeSniffer.


phpcs.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
```
Swangeon, line 1: Should Probably add that the user should download `php-codesniffer` to their machine in the README. Up to you on that as I think it could just be an optional requirement for the user.

Swangeon (changes requested):
Great work overall on this, just some small changes with adding a path and possibly add an optional requirement for the codesniffer itself.

phpcs.xml
```diff
@@ -0,0 +1,18 @@
...
+<ruleset name="BeaverHealth Code Style">
+  <description>PHP CodeSniffer rules for BeaverHealth project.</description>
+  <!-- Use PSR-12 Standard -->
+  <rule ref="PSR12"/>
+  <!-- Directories to include -->
+  <file>app/</file>
```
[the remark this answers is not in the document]
c-a-ray: I looked into adding `public/` to the PHPCS checks, but decided against it because:
1. The main file there (`index.php`) is Laravel's entry point file which intentionally mixes declarations and side effects by design, which would trigger unavoidable PSR warnings that we can't and shouldn't fix.
2. While there could theoretically be additional PHP files in `public/` in the future, it sounds like this is generally discouraged as a practice. The `public/` directory is meant primarily for the entry point and web-accessible assets, not application code. Any PHP logic would live in the appropriate directories that are already covered by the PHPCS rules (`app/`, `routes/`, etc.).

phpcs.xml
c-a-ray: Per our conversation, I created a `CONTRIBUTING.md` file with some information about this.

I kept the file pretty minimal since it was starting to get a little scope-creepy, but I also included some general information about setup, etc.

We can look at filling it out more in the future.

Swangeon: sounds good!

Swangeon (approved):

PR #60, opened 2025-02-02
Fleshed out Upload and Request Pages

BrynnMartina:
Upload:
-Added name and DOB inputs so that uploading a file is associated with a certain user
-Files are stored in storage/app/public/patient_records/[user_id]_[name]

Request:
-Keywords input added to showcase command injection
-Ex: ";whoami", "; ls" shows sensitive files
-Otherwise, correct info will print out all the files (unless you enter a keyword that filters the filenames). Clicking the links will download them

You have to pre-populate the files initially for the request page to show files


c-a-ray (changes requested):
Nice work. This is the most complicated feature we have, so great job getting it all working.

I left a few comments about how we could improve usability, but I would be ok with those suggestions not making it into v0.5. I would just ask that we talk about it and look at potentially making those changes for v0.6.

I would like to get the controllers cleaned up though. They are doing a lot more than handling HTTP requests, which among other issues would make testing very difficult.

PR #62, opened 2025-02-02
Converted Profile page to Change Password page 

alherrera-cs:
What’s Changed:

   - Refactored Profile Page → Now dedicated to Change Password instead.
   - Removed Unused Forms → Deleted partials folder with delete user and update profile forms since they’re no longer needed.
   - Updated Sidebar Navigation → Adjusted links to reflect the new Change Password page.
   - Updated Routes & Controller → Adjusted logic to handle password changes properly.
   - Moved Business Logic to a Service → Created ChangePasswordService.php to handle password updates.

Additional Changes (02/07):
- Removed unused profile methods in ProfileController and renamed change password view method.
- Updated ChangePasswordRequest to redirect back with validation errors.
- Combined password update logic in ChangePasswordService.
- Revised Blade view (change-password) to extend the shared layout and use inline error handling.
- Merged latest master changes for route and configuration consistency.


app/Http/Controllers/ProfileController.php
```diff
@@ -11,50 +13,51 @@
...
     {
         return view('profile.edit', [
             'user' => $request->user(),
         ]);
     }
 
-    /**
-     * Update the user's profile information.
-     */
     public function update(ProfileUpdateRequest $request): RedirectResponse
```
c-a-ray, line 30: This method isn't being used so we should delete it. Also, wherever you see reference to `email`, it's something that either needs to be deleted or changed, since we're not using `email`.

app/Http/Controllers/ProfileController.php
```diff
@@ -11,50 +13,51 @@
...
-     * Display the user's profile form.
-     */
+    protected $changePasswordService;
+
+    public function __construct(ChangePasswordService $changePasswordService)
+    {
+        $this->changePasswordService = $changePasswordService;
+    }
+
     public function edit(Request $request): View
```
c-a-ray, line 23: This can be deleted, I think.

app/Http/Controllers/ProfileController.php
```diff
@@ -11,50 +13,51 @@
...
+        $user->save();
 
         return Redirect::route('profile.edit')->with('status', 'profile-updated');
     }
 
-    /**
-     * Delete the user's account.
-     */
-    public function destroy(Request $request): RedirectResponse
+    public function showChangePasswordForm(): View
```
c-a-ray, line 44: To stay consistent with all of the other controllers, this should be called `index`. It's not a bad method name, but currently every other method that shows the main view is called `index`, so this is a little out of place.

app/Http/Controllers/ProfileController.php
```diff
@@ -11,50 +13,51 @@
...
-        $request->validateWithBag('userDeletion', [
-            'password' => ['required', 'current_password'],
-        ]);
-
-        $user = $request->user();
+        return view('profile.change-password');
+    }
 
-        Auth::logout();
+    public function changePassword(ChangePasswordRequest $request): RedirectResponse
```
c-a-ray, line 49: Nice and clean. Well done.

app/Services/ChangePasswordService.php
```diff
@@ -0,0 +1,20 @@
...
+class ChangePasswordService
+{
+    public function updatePassword(User $user, ChangePasswordRequest $request): bool
+    {
+        if (!Hash::check($request->input('current_password'), $user->password)) {
+            return false;
+        }
+
+        $user->password = Hash::make($request->input('password'));
+        return $user->save();
```
c-a-ray, line 18: Consider combining these two lines:
```php
return $user->update(['password' => Hash::make($request->input('password'))]);
```

resources/views/components/sidebar.blade.php
```diff
@@ -35,9 +35,8 @@
             <!-- Vulnerability Toggles Link -->
             <a href="{{ route('vulnerability_toggles') }}" class="block py-2 px-4 rounded hover:bg-gray-600">Vulnerability Toggles</a>
 
-            <!-- Change Password Link -->
-            <a href="{{ route('profile.edit') }}" class="block py-2 px-4 rounded hover:bg-gray-600">Profile</a>
-
+            <a href="{{ route('profile.change-password') }}" class="block py-2 px-4 rounded hover:bg-gray-700">Change Password</a>
```
c-a-ray, line 38: Can you add the comment above this back in? Just that every other link has one.

```
<!-- Change Password Link -->
```

resources/views/profile/change-password.blade.php
```diff
@@ -0,0 +1,84 @@
+<!DOCTYPE html>
```
c-a-ray, line 1: This Blade file is defining a full HTML structure rather than extending the existing layout. To maintain consistency and avoid duplication, this file should extend `app.blade.php`.

We use `<x-app-layout>` (defined in `resources/views/layouts/app.blade.php`) so that shared elements (sidebar, scripts, styles, etc.) are consistently applied across all pages. Also makes future updates easier, since changes to the layout will automatically reflect on all pages using it.

To fix this, replace the full HTML structure with something like this:
```blade
<x-app-layout>
    <x-slot name="header">
        <h2 class="font-semibold text-xl text-gray-800 dark:text-gray-200 leading-tight">
            {{ __('Update Password') }}
        </h2>
    </x-slot>

    <div class="p-8">
        // Page content here
    </div>
</x-app-layout>
```

routes/web.php
```diff
@@ -29,9 +29,8 @@ function () {
     function () {
```
c-a-ray, line 29: Looks like we still need to remove the `/confirm-password` and `/password` routes. Also I think we can delete `NewPasswordController`, `PasswordController`, and `PasswordResetLinkController`.

app/Http/Requests/ChangePasswordRequest.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        return [
+            'current_password' => 'required|string',
+            'password' => 'required|string|min:8|confirmed',
+        ];
+    }
+
+    protected function failedValidation(Validator $validator): void
+    {
+        throw new HttpResponseException(
```
c-a-ray, line 27: This is returning a JSON response rather than redirecting back with validation errors. Since the change password form expects a redirect with error messages, this approach results in raw JSON being displayed instead of showing validation errors in the UI.

To fix this, you can use a `ValidationException` instead of an `HttpResponseException`:
```php
use Illuminate\Validation\ValidationException;
...
    protected function failedValidation(Validator $validator): void
    {
        throw new ValidationException(
            $validator,
            redirect()->back()->withErrors($validator)->withInput()
        );
    }
```

You also need to make some changes in the blade file. I'll make a comment there explaining what you need to do.

resources/views/profile/change-password.blade.php
```diff
@@ -0,0 +1,84 @@
...
+                            </p>
+                        </header>
+
+                        <form method="post" action="{{ route('profile.change-password.update') }}" class="mt-6 space-y-6">
+                            @csrf
+
+                            <div>
+                                <x-input-label for="current_password" :value="__('Current Password')" />
+                                <x-text-input id="current_password" name="current_password" type="password" class="mt-1 block w-full" autocomplete="current-password" />
+                                <x-input-error :messages="$errors->changePassword->get('current_password')" class="mt-2" />
```
c-a-ray, line 50: `<x-input-error>` doesn't work here

After you make the change I talk about in the `ChangePasswordRequest` class, you'll need to replace each of the instances of `<x-input-error>` with something like this:
```blade
@error('current_password')
    <p class="text-red-500 text-sm mt-1">{{ $message }}</p>
@enderror
```

c-a-ray (changes requested):
Nice work. This is coming along well, but there are a few things that need to be fixed.

Biggest thing is that the validation error messages aren't working. Currently, if you click the "Save" button with invalid inputs, you get raw JSON like this:

image

I explain why this is happening in another comment, as well as how to fix it.

Next biggest thing is that the Blade file doesn't use the existing layout and duplicates the whole HTML structure of the app. I explain how to use the existing layout in another comment.

There are also a handful of little things, and a few routes and controllers that still need to be deleted.

All of this should be pretty easy/quick, so it's definitely possible to get this merged today if you're available to commit fixes.

app/Http/Controllers/ProfileController.php
```diff
@@ -2,59 +2,39 @@
...
-            'user' => $request->user(),
-        ]);
+        $this->changePasswordService = $changePasswordService;
     }
 
-    /**
-     * Update the user's profile information.
-     */
-    public function update(ProfileUpdateRequest $request): RedirectResponse
+    public function index(Request $request): View
```
c-a-ray, line 21: Unused parameter

resources/views/profile/change-password.blade.php
```diff
@@ -0,0 +1,64 @@
...
+            </header>
+
+            <form method="post" action="{{ route('profile.change-password.update') }}" class="mt-6 space-y-6">
+                @csrf
+
+                <div>
+                    <x-input-label for="current_password" :value="__('Current Password')" />
+                    <x-text-input id="current_password" name="current_password" type="password" class="mt-1 block w-full" autocomplete="current-password" />
+                    @error('current_password')
+                        <p class="text-red-500 text-sm mt-1">{{ $message }}</p>
```
c-a-ray, line 26: Not totally sure what's going on here, but the error text is super hard to read when system theme is dark. I see you have `text-red-500`, but it's showing up as black for some reason. Would you mind looking into that?

image

c-a-ray (changes requested):
Good stuff, coming together. Very close, just a few small things.

Looks like there was one [comment](https://github.com/BeaverHealth-Vulnerable-Web-App/BeaverHealth-Vulnerable-Web-App/pull/62/files#r1938631922) from the previous round of review that you didn't get to.

Also, we are now using PHP_CodeSniffer to make sure all of our code adheres to PSR-12. Looks like there are some unresolved codesniffer issues here. Check out the new `CONTRIBUTING.md` file to see how to use it.

Other than that, I think there are just two additional comments about small things.

resources/js/vulnerability_toggles.js
```diff
@@ -1,6 +1,7 @@
 let notificationTimeoutID;
 
-function showNotification(message, success) {
+function showNotification(message, success)
```
c-a-ray, line 3: Curious why you're reformatting JavaScript files? Codesniffer shouldn't have anything to say about these. We don't have a formatter for JS like we do for PHP. But it raises a good question about whether we should introduce a JS formatter like Prettier or ESLint. Thoughts?

tailwind.config.js
```diff
@@ -7,15 +7,15 @@ export default {
         './vendor/laravel/framework/src/Illuminate/Pagination/resources/views/*.blade.php',
         './storage/framework/views/*.php',
         './resources/views/**/*.blade.php',
-    ],
+        ],
```
c-a-ray, line 10: The formatting changes to the other JS files look ok to me, but this doesn't look quite right. Shouldn't it be more like this?

```javascript
export default {
    content: [
        './vendor/laravel/framework/src/Illuminate/Pagination/resources/views/*.blade.php',
        './storage/framework/views/*.php',
        './resources/views/**/*.blade.php',
    ],
    theme: {
        extend: {
            fontFamily: {
                sans: ['Figtree', ...defaultTheme.fontFamily.sans],
            },
        },
    },
    plugins: [forms],
};
```

c-a-ray (commented):
Nice work. This looks pretty good, so I'd be willing to merge it as-is, but there are a couple super small things that it would be nice to do if you have time today.

- There's still an [unresolved comment](https://github.com/BeaverHealth-Vulnerable-Web-App/BeaverHealth-Vulnerable-Web-App/pull/62/files#r1938631922) from my first review about removing the unused routes and controllers related to password/profile updates. I think we can remove the following:
  - In `web.php`:
    - `/confirm-password` route
    - `/password` route
  - `NewPasswordController`
  - `PasswordController`
  - `PasswordResetLinkController`
- Not sure why you reformatted those JavaScript files, but I'm cool with it. The only thing that doesn't look quite right is `tailwind.config.js`. I made a comment on that file with more details.

Let me know if you can delete those routes and controllers and reformat that one file today. If not, I'll approve as-is and we can fix them next week.

resources/js/vulnerability_toggles.js
alherrera-cs: I didn’t intentionally change the JS files

alherrera-cs: Not sure exactly what happened there; I’m reverting the JS files back to their previous state. Yeah we should probably look into adding a dedicated JS formatter like Prettier or ESLint to help maintain consistency

c-a-ray (approved):
Nice!

PR #64, opened 2025-02-02
Fix setup process and automate deployment

c-a-ray:
This PR fixes our current setup process.

Currently, we run `composer install` directly on the host machine. This approach has multiple issues:
- Requires developers to have PHP and Composer installed locally.
- Can lead to inconsistencies due to different PHP versions and dependencies.

The main thing this PR does is add a `setup.sh` script that will allow both users and developers to deploy the application with a single command, without having to install application dependencies on their local machine.

With this change, new users only need to:
1. Make sure Docker/Docker Compose is available on the host
2. Clone the repo
3. Execute `setup.sh --fresh`

The script is intended to be multi-purpose. I.e., users/developers can use this for an initial deployment as well as subsequent deployments.

Available flags are:
- `-f | --fresh`: intended to be used for initial setup or for doing a total refresh (installs deps, rebuilds app, wipes/migrates/seeds DB, clears caches)
- `-i | --interactive`: provides interactive prompts to choose which actions are taken
- `-h | --help`: show the help message

Either `--fresh` or `--interactive` must be used. Options are mutually exclusive.


setup.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+
+set -e
+
+echo "Installing Laravel dependencies inside of a container..."
```
Swangeon, line 5: would maybe add a bit more distinction when the script uses echo like adding ANSI escape code colors but is just a personal thing. Not gonna hold the PR up on it.

Swangeon (approved):
Great job on simplifying the process, not much to really say about this as it is fairly small and I am fine with it being merged with master soon. Just made one comment but it's no big deal.

setup.sh
c-a-ray: Good suggestion. Updated the script to add colors.

Also added flags to the script to make it more multi-purpose.

setup.sh
```diff
@@ -0,0 +1,129 @@
...
+if [[ "$INSTALL_DEPS" == true ]]; then
+    echo -e "${BLUE}Installing Laravel dependencies inside of a container...${NO_COLOR}"
+    docker run --rm \
+        -u "$(id -u):$(id -g)" \
+        -v "$(pwd)":/var/www/html \
+        -w /var/www/html \
+        laravelsail/php82-composer:latest \
+        composer install --optimize-autoloader
+fi
+
```
Swangeon, line 87: Ideally I would like to have some level of error handling here just in case teh user runs the script without doing the `-f/--fresh` option first. If the `vendor` directory can't be found and the `-f/--fresh` option is not true, then the application should throw an error like
```bash
if [[ ! -d 'vendor/' && "$FRESH_INSTALL" != true ]]; then
    echo -e "${RED}Sail was not found, run './setup.sh --fresh' to build it.${NO_COLOR}" && exit 1
fi
```
Although you might be able to omit the ` && "$FRESH_INSTALL" != true` part as the install dependencies runs before this.

Swangeon (changes requested):
Great work with adding the colors and the options to it! Just one error handling thing that I would like to add as a just in case the script is ran without the `-f/--fresh` options first.

setup.sh
c-a-ray: Added a check to ensure `vendor/` exists before any `sail` commands are run. Also added a preflight check to make sure Docker is installed and running.

Script has also changed a bit in other ways since your last review.

Now, the only flags are `--fresh` and `--help`. If you don't use `--fresh`, there are interactive prompts to determine which actions to take. Makes it a little nicer to use and gives more fine-grained control.

Also did some general cleanup and put everything into functions to make it easier to read.

database/migrations/2024_11_19_153826_create_user_table.php
```diff
@@ -6,7 +6,7 @@
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\Schema;
 
-class CreateUserTable extends Migration
+return new class extends Migration
```
c-a-ray, line 9: Changed migrations from named classes to anonymous classes to resolve a migration error. Randomly started getting error `Cannot declare class Database\Migrations\CreateUserTable, because the name is already in use`. The error was happening despite having only one `CreateUserTable` class in the codebase. Using anonymous classes sidesteps the issue by avoiding migration class name declarations entirely. Also, this matches how Laravel's own migrations are structured -- e.g., the cache table migration that came with Laravel uses anonymous classes by default.

Swangeon (approved):

PR #68, opened 2025-02-08
Fully Developed Add & Request Pages

BrynnMartina:
### What it Does

Add Records:
- Dropdown to select patient name and their associated SSN
- File upload that is associated with the chosen patient (vulnerable)
- Patients that do not exist and/or blank inputs will not be accepted

Request Records:
- Dropdown to select patient name and their associated SSN
- Filter keyword box to filter document names (vulnerable)
- Patients that do not exist and/or blank inputs will not be accepted

02/16/25 Updates:
-Dropdown will open when you click on the text box, instead of having to start typing.
-Request and Add form boxes have equal width
-Service files merged into one "PatientRecordService.php"

### Filesystem

app/
│─ Http/
│     ├─ Controllers/
│     │   ├─ Records/
│     │   │   ├── AddRecordsController.php (Handles record additions & file uploads)
│     │   │   ├── RequestRecordsController.php (Handles searching & retrieving patient records)
│     ├─ Requests/
│     │   ├─ Records/
│     │   │   ├── StoreRecordRequest.php (Validates patient ID & file upload request)
│     │   │   ├── PatientSearchRequest.php (Validates patient ID & keyword for searching)
│─ Services/
│     ├─ PatientRecordService.php
│─ Presenters/
│     ├── FilePresenter.php (Formats file download links for display)

Additional Files:
patient-dropdown.blade.php (universal patient dropdown formatting)
add.blade.php (add page formatting)
request.blade.php (request page formatting)


c-a-ray (changes requested):
Hell yeah dude. Looking great. I think we're really close on this, just a couple things to consider.

Basically two main things, and one nice-to-have. 
1. I think it might make more sense to have a single `PatientRecordService` (instead of four different ones). There's another comment that talks more about this.
2. The "Request Records" component is much wider than the "Add Records" component.
i.e.
image
vs.
image
Seems like these things should be roughly the same size.


**Nice-to-have**
Haven't at all looked into how this would be done, so don't worry about it if it's a ton of work or something, but it would be nice to have the patient dropdown do a few things (in order of nice-to-haveness):
- Highlight the entry being hovered over
- Open an unfiltered list of patients when clicking into the input (as opposed to having to type a letter before seeing anything)
- Be able to use arrow keys to select an entry

We could totally do this in another PR though. No reason to hold up this PR just for that.

---

Also, not sure if you saw my PHP_CodeSniffer PR, but we now have a PHP_CodeSniffer config (`phpcs.xml`) so we can keep the codebase consistent with PSR-12 standards. Looks like this has some PSR-12 issues.

The new `CONTRIBUTING.md` file has some info on how to run PHP_CodeSniffer. Basically you just need to run `phpcbf` (which automatically fixes styling/formatting issues) and then run `phpcs` to identify issues that couldn't be fixed automatically.

PR #82, opened 2025-02-11
Update CI

c-a-ray:
This PR updates the CI workflow to run our `PHPUnit` tests and `phpcs`.

A few incidentals were needed to get this to work:
- Adding `PHP_CodeSniffer` as a dependency
- A new `docker-compose.ci.yaml` that doesn't mount volumes

This branch is based off of `cr-fixDeployment` because that branch fixes a migration issue that is required to get this to work. We will need to get PR #64 merged before this.

I explored splitting up the workflow into multiple jobs, but this introduces complications for no real benefit. Since all tasks (migrations, tests, and PHP_CodeSniffer) depend on the same containerized environment, it makes more sense to keep these steps in the same job. This also simplifies the pipeline and makes debugging easier by centralizing logs and errors.


.dockerignore
```diff
@@ -2,3 +2,5 @@ public/storage
 public/build
 vendor
 node_modules
+.env
```
c-a-ray, line 5: Excluded so the correct `.env` file can be mounted depending on the deployment environment

.env
```diff
@@ -23,7 +23,7 @@ LOG_DEPRECATIONS_CHANNEL=null
 LOG_LEVEL=debug
 
 DB_CONNECTION=mysql
-DB_HOST=127.0.0.1
+DB_HOST=db
```
c-a-ray, line 26: This change was needed to get the database connection working in the CI workflow. It turns out to also work locally. Initially I thought I would need to create a separate `.env.ci` with this change, but it turns out we can just make this change in the existing `.env` and everything is gravy.

.github/workflows/ci.yaml
```diff
@@ -10,24 +10,39 @@ on:
   workflow_dispatch:
 
 jobs:
-  laravel-tests:
+  beaverhealth-ci:
```
c-a-ray, line 13: As mentioned in the PR description, I tried to split this up into separate jobs, but that turns out to be problematic. Since our CI pipeline is so simple, and everything depends on the same container setup process, there's really no benefit to splitting this up into multiple jobs. Keeping it all in one job also makes debugging issues easier.

composer.json
```diff
@@ -19,7 +19,8 @@
         "mockery/mockery": "^1.6",
         "nunomaduro/collision": "^8.1",
         "pestphp/pest": "^3.5",
-        "pestphp/pest-plugin-laravel": "^3.0"
+        "pestphp/pest-plugin-laravel": "^3.0",
+        "squizlabs/php_codesniffer": "^3.11"
```
c-a-ray, line 23: Had to add `php_codesniffer` here so it's available in the container for CI to run

PR #85, opened 2025-02-17
Feedback Page Refactor

Swangeon:
The Patient Feedback page was one of the first pages we implemented and a lot of change in design decisions have been made since then and the page now needs to be updated to reflect those changes. 

Frontend
- Adds the dark theming to the page 
- Adds the new Patient Dropdown component for easier use of the page

Backend
- Cleaned up a lot of the code in the controller 
- Adds 2 new request classes to validate the patient feedback and the feedback search form data 
- Separated the main logic for these functionalities into a service class

### DIFF
--- a/tests/Feature/FeedbackFeatureTest.php
+++ b/tests/Feature/FeedbackFeatureTest.php
@@ -9,12 +9,14 @@ use Tests\TestCase;
 class FeedbackFeatureTest extends TestCase
 {
     protected $user;
+    protected $patient;
 
     public function setUp(): void
     {
         parent::setUp();
         $this->user = User::factory()->create();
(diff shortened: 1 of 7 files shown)


tests/Feature/FeedbackFeatureTest.php
```diff
@@ -2,74 +2,81 @@
```
c-a-ray, line 2: Nice

c-a-ray (changes requested):
Very nice. 

Just found two very minor things:
- Some unused imports in `FeedbackController`
- One little styling thing in the feedback view. Easy fix.

c-a-ray (approved):
Good stuff!

PR #93, opened 2025-02-19
Add feature tests for change password page

alherrera-cs:
This PR adds new feature tests for the change password page. The tests cover a variety of scenarios such as:

    - Making sure the page loads correctly.
    - Successfully updating the password when given the correct current password.
    - Handling cases where the current password is wrong.
    - Catching issues like password confirmation mismatches and too-short new passwords.
    - Protecting against basic SQL injection attempts.
    - Testing for missing required fields.

### DIFF
--- a/tests/Feature/ChangePassFeatureTest.php
+++ b/tests/Feature/ChangePassFeatureTest.php
@@ -1,0 +1,195 @@
+<?php
+
+namespace Tests\Feature;
+
+use App\Models\User;
+use Illuminate\Support\Facades\Hash;
+use Tests\TestCase;
+
+class ChangePasswordFeatureTest extends TestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->startSession();
+    }
+
+    protected function createTestUser(string $password = 'password123'): User
+    {
+        return User::factory()->create([
+            'password' => bcrypt($password),
+        ]);
+    }
+
+    public function testChangePasswordPageAccessible(): void
+    {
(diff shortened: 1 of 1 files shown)


tests/Feature/ChangePassFeatureTest.php
```diff
@@ -0,0 +1,195 @@
+<?php
```
c-a-ray, line 1: Can we rename this to `ChangePasswordFeatureTest.php`?

tests/Feature/ChangePassFeatureTest.php
```diff
@@ -0,0 +1,195 @@
...
+
+        $this->assertTrue(Hash::check('newpassword123', $user->fresh()->password));
+    }
+
+    public function testChangePasswordFailsWithIncorrectCurrentPassword(): void
+    {
+        $user = $this->createTestUser('oldpassword');
+        $this->actingAs($user);
+
+        $response = $this->withHeaders(['referer' => route('profile.change-password')])
```
c-a-ray, line 56: IMO these are a lot easier to read if formatted like this:
```php
        $response = $this->postWithCsrf(
            route('profile.change-password.update'),
            [
                'current_password'      => 'oldpassword',
                'password'              => 'newpassword123',
                'password_confirmation' => 'newpassword123',
            ]
        );
```

That's how they are written in every other feature test file, probably good to stay consistent.

c-a-ray (commented):
Nice work. Overall, a pretty solid test suite. Just a few suggestions.

PR #94, opened 2025-02-20
Add pre-commit and pre-push hooks

Swangeon:
The pre-commit hook will run `phpcs` to find any style violations, and the pre-push hook will run the testing suite through `./vendor/bin/sail artisan test` to make sure the app is still behaving appropriately. These git hooks are installed on a developer's machine through running `composer install` or running the setup script with the `--fresh` option. 

The main reason I decided to go the path of creating these hooks through `composer` was because it cuts down on making 2 new bash scripts that effectively run one command and a bash script that copies them into the developer's `.git` directory as `brainmaestro/composer-git-hooks` does all of that heavy lifting after an installation of all dependencies. Along with that, multiple resources online were saying that if you are developing a project that is using `composer` then using the module is the most standard way to transfer git hooks to all developers on the project.

Another thing to note, `phpcs` won't provide any output if it finds nothing wrong with the code styling when it is running in a script (or if it's running in `sh`, I don't quite know) but will provide output and stop the commit if you do have any style violations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ yarn-error.log
 /.idea
 /.vscode
 /.zed
+cghooks.lock
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,10 @@
     "name": "laravel/laravel",
     "type": "project",
     "description": "The skeleton application for the Laravel framework.",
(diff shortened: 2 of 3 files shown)


c-a-ray (approved):
Nice

PR #95, opened 2025-02-22
UI updates

c-a-ray:
This PR fixes a handful of UI things that have been bugging me. Its goal is to make the app look cleaner and more consistent. Main things it does:
- Adds consistent dark mode styling across the entire app
- Adds a shared component for status messages
- Makes it so updating a role in the Admin page dynamically updates the sidebar, without a page refresh

Some small incidentals it also takes care of:
- Aligns the buttons on the login page
- Updates the tab title to "BeaverHealth" (previously read "Laravel")
- Includes the uploaded filename in the Add Records status message
- Removes unused files (mostly unused Blade templates)

UI updates:

Login page buttons (feel like this could maybe be better, but haven't found anything I really like)
image

- Sidebar updates
image

- Admin page table
image
image

- Vulnerability Toggles table
image
image

- Patient Info table
image
image

- Patient Feedback dark mode
image
image

- Container for Change Password form
image
image

- Consistent status message popups across the app
image
image
image
image


resources/js/vulnerability_toggles.js
```diff
@@ -1,61 +1,70 @@
...
-    clearTimeout(notificationTimeoutID);
-    notification.classList.remove('hidden');
-    notificationText.textContent = message;
-    if (success === 0) {
-        notification.querySelector('div').className =
-            'px-4 py-2 rounded-lg shadow-lg text-center min-w-[200px] bg-green-100 text-green-800 dark:bg-green-800 dark:text-green-100';
-    } else {
-        notification.querySelector('div').className =
-            'px-4 py-2 rounded-lg shadow-lg text-center min-w-[200px] bg-red-100 text-red-800 dark:bg-red-800 dark:text-red-100';
+document.addEventListener('DOMContentLoaded', function() {
```
c-a-ray, line 1: Initially edited this file to move the notification to the top right, improve dark-mode support, and fix the error styles not being applied, but then figured I might as well do a proper refactor by:
- Centralizing notification styles in a config object
- Adding proper timeout/cleanup management 
- Adding smooth fade animations
- Improving error handling and state recovery
- Converting to async/await pattern

resources/views/profile/change-password.blade.php
```diff
@@ -5,60 +5,65 @@
...
-        <section>
-            <header>
-                <h2 class="text-lg font-medium text-gray-900 dark:text-gray-100">
-                    {{ __('Update Password') }}
-                </h2>
-                <p class="mt-1 text-sm text-gray-600 dark:text-gray-400">
-                    {{ __('Ensure your password is at least 8 characters long.') }}
-                </p>
-            </header>
+    <div class="py-12">
```
c-a-ray, line 8: Adds a container for the form to make it consistent with the rest of the app.

BrynnMartina: Another potential area we could universalize. You could create a container component 

image

which would allow you to replace `<div class="max-w-7xl mx-auto sm:px-6 lg:px-8">` with `<x-container>`

I don't see this as being as big of a priority as other elements, but you get the point.

BrynnMartina (commented):
Overall, these modifications look great and I don't see anything holding it back from being merged. The modifications make the website look much more cohesive. UI has been on the back burner in the grand scheme of things, and I think there's an opportunity to talk more about this at our weekly meeting. For the future there may be some perks towards making some of these UI features more universally accessible. And this job does not necessarily have to fall to you.

For example, success and error messages are not currently consistent between all pages. We could utilize resources/views/components more, similar to the patient-dropdown.blade.php file made for the dropdown listing patient names and SSN.

We could make a new file, like notification.blade.php:

Screenshot from 2025-02-24 01-52-13

Add a component in add.blade.php:

image

And have the controllers handle the rest (which some pages already do currently)
`return back()->with('success', 'File uploaded successfully!');`


Unrelated:
On the login page the left side of the "Log In" button does not align with the "Username" and "Password" text boxes. To align it, you can change `ms-3` to `ml-0`. Just a personal preference of mine.

Login.blade.php:
image

c-a-ray:
> For example, success and error messages are not currently consistent between all pages. We could utilize resources/views/components more, similar to the patient-dropdown.blade.php file made for the dropdown listing patient names and SSN.
> 
> We could make a new file, like notification.blade.php:

That's so funny -- I didn't realize you had reviewed my PR, but I made exactly this change. Added a `status-message.blade.php` Blade template and a `StatusMessage` component class, and updated all of the pages to use it.

c-a-ray:
> Unrelated: On the login page the left side of the "Log In" button does not align with the "Username" and "Password" text boxes. To align it, you can change `ms-3` to `ml-0`. Just a personal preference of mine.

I like your thinking here. You're right that the Login page buttons could look better. I was playing around with it and landed on this:
image

What do you think?

Using `ml-0` does look better than `ms-3`, for sure. But then I was trying to make it so the "CREATE AN ACCOUNT" button and the "LOG IN" button were lined up horizontally and it didn't look quite right. Centering them seemed to look the best to me, but I could definitely be convinced otherwise.

app/Http/Requests/ProfileUpdateRequest.php
```diff
@@ -1,28 +0,0 @@
-<?php
```
c-a-ray, line 1: File is unused

resources/views/profile/change-password.blade.php
c-a-ray: I looked into creating a container component but it turned out to actually be kind of difficult. There are a lot of styling variations for containers between pages, so a single reusable component would need a lot of parameters or overrides. Just seems like it's more trouble than it's worth.

Not opposed to the idea if you want to give it a shot though.

app/Http/Controllers/PasswordController.php
```diff
@@ -8,7 +8,7 @@
 use Illuminate\Support\Facades\Auth;
 use Illuminate\View\View;
 
-class ProfileController extends Controller
+class PasswordController extends Controller
```
c-a-ray, line 11: Just seems more intuitive, since all it does is change the password

app/Http/Requests/ProfileUpdateRequest.php
c-a-ray, line 1: Not used

resources/js/utils.js
```diff
@@ -0,0 +1,31 @@
+export function showStatusMessage(message, type = 'success') {
```
c-a-ray, line 1: This `showStatusMessage` function is needed for pages that do client-side updates (AJAX). The status message component can't be used in those cases because it is rendered server-side. Currently, that's only the Vulnerability Toggles page, but I'm working on a separate PR that uses it on the Admin page. Extracted it into `utils.js` so it can be shared.

resources/views/admin.blade.php
```diff
@@ -4,83 +4,45 @@
...
+                                </td>
+                            </x-table-row>
+                        @endforeach
+                    </x-table>
                 </div>
             </div>
         </div>
     </div>
-
-    <!-- JavaScript to handle AJAX requests -->
```
c-a-ray, line 47: Moved to `resources/js/admin.js`. Note that any changes to those files require re-running `npm run build` in the app container.

resources/views/records/add.blade.php
```diff
@@ -1,7 +1,7 @@
 <x-app-layout>
     <x-slot name="header">
         <h2 class="font-semibold text-xl text-gray-800 dark:text-gray-200 leading-tight">
-            {{ __('Upload Patient Records') }}
+            {{ __('Add Medical Records') }}
```
c-a-ray, line 4: Updated the page titles to be consistent with the sidebar, etc. We tend to use slightly different names for things a lot, seems better to have more consistency with what we call things.

app/Http/Controllers/AddRecordsController.php
```diff
@@ -23,11 +23,14 @@ public function index()
 
     public function upload(StoreRecordRequest $request)
     {
+        $file = $request->file('medical_record');
```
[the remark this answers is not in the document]
alherrera-cs: good catch!

PR #96, opened 2025-02-23
Bd add and request records

BrynnMartina:
Added tests for add records and request records pages.
Error handling, such as making sure patient names are filled out correctly (or not correctly), are handled on the front end currently.

Add records:
-Checks if page loads
-Checks if files load correctly
-Checks if entry in database is created correctly

Request records:
-Checks if page loads
-Checks if patient info is retrieved correctly
-Checks if "No files found" prints when a patient with no files is searched
-Checks if files are printed when a patient with files is searched
-Checks if keyword search works correctly
-Checks if files are downloaded correctly


tests/Feature/RequestRecordsFeatureTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $response = $this->post(route('records.search'), [
+            '_token'     => csrf_token(),
+            'patient_id' => $this->patient->patient_id,
+            'keyword'    => '2',
+        ]);
+
+        $response->assertRedirect(route('records.request'));
+
+        $patientFiles = session('patient_files');
+        $this->assertIsString($patientFiles, 'Expected patient_files to be a string');
```
c-a-ray, line 125: Good test, but it took me a few minutes to figure out what the point of this assertion is. Might be a good place for a comment.

c-a-ray (changes requested):
Good work! Sorry for all of the comments -- just a lot of small things add up. Overall, it feels a bit messy in places, but everything should be pretty easy to clean up.

One thing to note: you used the same branch as your previous PR. Definitely best to create a new branch for each PR to keep things clean and avoid potential issues.

PR #98, opened 2025-02-27
Secure access control

c-a-ray:
This PR changes the `idor_on` toggle to `bac_on` and implements a secure version of the access control system. With the BAC vulnerability toggled off, the user is unable to manipulate the URL to navigate to pages for which they do not have the appropriate role. This is primarily done via a custom middleware `CheckPermissionMiddleware`.

If BAC is enabled, the user can update roles even if they do not have the admin role. If BAC is disabled and the user attempts to access a protected page via URL manipulation, they are redirected to the dashboard and shown an "Access denied" error message. If BAC is disabled, the user is on the admin page, and removes their own admin role, the role update is blocked and an "Insufficient permission" status message is shown.

This PR also does a bit of cleanup of files related to the admin page, and updates the access control feature tests. Because of the addition of the ability to turn BAC on and off, the number of test cases became very large, so I refactored most of the access control feature tests to use data providers. It should still cover all of the same test cases (and more), even though it's fewer lines of code.


tests/Feature/AccessControlFeatureTest.php
```diff
@@ -16,124 +17,280 @@ protected function createUserWithRoles($roles = [])
...
-                    'role' => $role,
-                    'value' => true
-                ]
-            )
-            ->assertOk()
-            ->assertJson(['success' => true]);
+    #[DataProvider('updateRoleDataProvider')]
+    public function testUpdateRoleValidation(array $requestData, array $expectedErrors)
+    {
+        if (!isset($requestData['user_id'])) {
```
c-a-ray, line 264: Because the data provider is static, you can't use `$this->createUser()->user_id` to set the user ID, so it has to happen here.

app/Http/Controllers/AddRecordsController.php
```diff
@@ -31,6 +31,11 @@ public function upload(StoreRecordRequest $request)
             $file
         );
 
-        return back()->with('success', "File '{$filename}' uploaded successfully!");
+        session()->flash('records-status', [
```
c-a-ray, line 34: Because I added a global status message for the "Access denied" message, I needed to be able to differentiate between where the statuses come from in the session. To achieve this, I updated all places where status messages are set to use `flash()` with specific session keys (i.e., `access-status`, `change-password-status`, `feedback-status`, and `records-status`). This makes sure that page-specific status messages don't conflict with the global access control status message.

app/Http/Middleware/CheckPermissionMiddleware.php
```diff
@@ -0,0 +1,55 @@
...
+        if (!$user->bac_on && $isProtectedRoute) {
+            $requiredPermission = $routePermissionMap[$routeName];
+            if (!$user->{$requiredPermission}) {
+                session()->flash('access-status', [
+                    'type' => 'error',
+                    'message' => 'Access denied: You do not have permission to view this page.'
+                ]);
+
+                // Add timestamp parameter to force refresh on browser back button
+                return redirect()->route('dashboard', ['_refresh' => time()]);
```
BrynnMartina, line 41: Great solution!

resources/js/admin.js
```diff
@@ -1,67 +1,95 @@
...
-                console.error('There was a problem:', error);
+        if (result.success) {
+            // Only update sidebar if user is changing their own role
+            if (targetUserId === window.currentUserId) {
+                await updateSidebar();
             }
-        });
-    });
+        } else {
+            revertCheckbox(checkbox, originalState, result.message || 'Failed to update permissions');
```
BrynnMartina, line 29: I like the specification with the error messages

BrynnMartina (approved):
Great work. I've done everything in my human power to test the BAC security and it's solid. Thanks for re-specifying the type of vulnerability from IDOR to BAC, too. And as always, the little details with making the success/error messages really refines the website.

With my rudimentarily trained eyes I don't see any structural or stylized coding issues. In any case, I can vouch for the fact that BAC is secure and I have not run into any functional errors on my end, so I'm more than comfortable giving this PR my blessing to merge.

PR #99, opened 2025-03-02
Secure Patient Feedback

Swangeon:
Adds the secure and insecure versions for the storing (Stored XSS) and search (Reflective XSS) features of the Patient Feedback page.

When the user has the Stored XSS vulnerability toggle on the logic in the service class will not sanitize the user input when storing it in the database for the **Add Comment** part of the page to exploit a Stored XSS vulnerability when rendering the page with all the feedback and when the toggle is off it will sanitize the user input before storing it in the database. When the Reflective XSS toggle is turned on, the name you put in the **Search Comments** part of the page will not sanitize the name when rendering it back to the user and the name is sanitized when the toggle is off.

A lot of this mostly came down to just refactoring the Service and Controller class, as I didn't quite take all the business logic out of the controller when I refactored it last time for the searching comments functionality, specifically when getting the user input for the name.

NOTE: While Blade templates have a built-in way to handle sanitization of the data passed into them through `{{ }}`, this page is using `{!! !!}` to make sure that the insecure version works without having to create an entire other template for the secured version.


app/Services/FeedbackService.php
```diff
@@ -16,22 +16,42 @@ public function getFeedbackWithPatients()
         return ['feedback' => $feedback, 'patients' => $patients];
     }
 
-    public function storeFeedback(FeedbackCommentRequest $request)
+    public function storeFeedbackSecure(FeedbackCommentRequest $request)
```
c-a-ray, line 19: So I know I said we should probably have separate methods for secure and insecure, but looking at this now, I'm thinking I was mistaken. I didn't realize the two methods would be exactly the same with one extra line in the secure version to escape the input. Dumb suggestion, totally my bad.

I know you don't need my help writing this, but just as an example for comparison purposes, `storeFeedback` could be as simple as this:
```php
public function storeFeedback(FeedbackCommentRequest $request, bool $xss_stored_on)
{
    $feedback = $request->input('feedback');
    if (!$xss_stored_on) {
        $feedback = e($feedback);
    }

    PatientFeedback::create([
        'patient_id' => $request->input('patient_id'),
        'feedback' => $feedback
    ]);
}
```

and basically same thing with `searchFeedback`. 

Up to you if you want to make this change. I think it's a lot cleaner, but it functions the same either way, so I wouldn't hold up your PR if you want to keep it like it is.

app/Services/FeedbackService.php
```diff
@@ -16,22 +16,42 @@ public function getFeedbackWithPatients()
         return ['feedback' => $feedback, 'patients' => $patients];
```
c-a-ray, line 16: We also need to escape the `$feedback` coming out the of database if `xss_stored_on` is false. E.g., if you turn on `xss_stored_on`, inject script into the database, and then turn off `xss_stored_on`, script in the database is still executed when visiting the Patient Feedback page.

Pretty simple fix though. I was able to fix it by just doing a `$feedback->each()` and escaping each feedback entry.

c-a-ray (changes requested):
Nice work. Only thing that _needs_ to happen is to escape feedback coming out of the database if `xss_stored_on` is false. I also suggest a slight refactor, but that's up to you.

app/Services/FeedbackService.php
```diff
@@ -16,22 +16,42 @@ public function getFeedbackWithPatients()
...
     }
 
-    public function storeFeedback(FeedbackCommentRequest $request)
+    public function storeFeedbackSecure(FeedbackCommentRequest $request)
+    {
+        $sanitizedFeedback = htmlspecialchars($request->input('feedback'), ENT_QUOTES, 'UTF-8');
```
c-a-ray, line 21: Forgot to put this in the review, but just wanted to make sure you knew that 

`htmlspecialchars($request->input('feedback'), ENT_QUOTES, 'UTF-8')`, for our purposes, is exactly the same as `e($request->input('feedback')`.

It's just more concise/readable and idiomatic. Does exactly the same thing though, so up to you.

app/Services/FeedbackService.php
```diff
@@ -6,32 +6,65 @@
...
+        if ($toggle) {
+            return $string;
+        } else {
+            return e($string);
+        }
+    }
+
+    private function sanitizeStoredFeedback(Collection &$feedback): void
+    {
+        for ($i = 0; $i < count($feedback); $i++) {
```
c-a-ray, line 24: > Suggestion

This can be done with a `foreach` (or `transform`):
```php
foreach ($feedback as $item) {
    if ($item->is_vulnerable) {
        $item->feedback = e($item->feedback);
    }
}
```
Benefits are improved readability, more efficient (doesn't call `count($feedback)` over and  over), less prone to errors, and more idiomatic.

Another way to handle this more functionally (I'm always a fan of the functional approach) is to use `map()`:
```php
private function sanitizeStoredFeedback(Collection $feedback): Collection
{
    return $feedback->map(function ($item) {
        if ($item->is_vulnerable) {
            $item->feedback = e($item->feedback);
        }
        return $item;
    });
}
```

app/Services/FeedbackService.php
```diff
@@ -6,32 +6,65 @@
...
+            if ($feedback[$i]->is_vulnerable == true) {
```
c-a-ray, line 25: > Suggestion

You don't need the `== true` since `is_vulnerable` is already a `bool`.

app/Services/FeedbackService.php
```diff
@@ -6,32 +6,65 @@
...
+        if (auth()->user()->xss_stored_on) {
+            return ['feedback' => $feedback, 'patients' => $patients];
+        }
+        $this->sanitizeStoredFeedback($feedback);
         return ['feedback' => $feedback, 'patients' => $patients];
     }
 
     public function storeFeedback(FeedbackCommentRequest $request)
     {
+        $feedback = $this->processString($request->input('feedback'), auth()->user()->xss_stored_on);
```
c-a-ray, line 44: > Suggestion

You could cut this method size in half with a little cleanup

- Since `auth()->user()->xss_stored_on` is called twice, might want to store it in a variable (or even better, pass it in from the controller)
- The `if`..`else` can be consolidated to one line with a ternary. 
- No need to store `$feedback` in a variable

```php
public function storeFeedback(FeedbackCommentRequest $request, bool $xss_stored_on)
{
    PatientFeedback::create([
        'patient_id' => $request->input('patient_id'),
        'feedback' => $this->prepareInput($request->input('feedback'), $xss_stored_on),
        'is_vulnerable' => $xss_stored_on ? true : false
    ]);
}
```

app/Services/FeedbackService.php
```diff
@@ -6,32 +6,65 @@
...
-        $name = $request->input('search_name');
+        $name = $this->processString($request->input('search_name'), auth()->user()->xss_reflected_on);
+
         $feedback = PatientFeedback::whereHas('patient', function ($query) use ($name) {
             $query->where('first_name', 'like', "%{$name}%")
                 ->orWhere('last_name', 'like', "%{$name}%");
         })->orderBy('created_at', 'desc')->get();
         $patients = Patient::all();
-        return ['feedback' => $feedback, 'patients' => $patients];
+        $this->sanitizeStoredFeedback($feedback);
```
c-a-ray, line 67: Wouldn't you only want to sanitize the feedback if `xss_stored_on` is false?

app/Services/FeedbackService.php
```diff
@@ -6,32 +6,65 @@
...
+        return [['feedback' => $feedback, 'patients' => $patients], $name];
```
c-a-ray, line 68: > Suggestion

Could get rid of the `$patients` assignment by just using `'patients' => Patient:all()` in the return

app/Services/FeedbackService.php
```diff
@@ -6,32 +6,65 @@
...
+                $feedback[$i]->feedback = e($feedback[$i]->feedback);
+            }
+        }
+    }
+
     public function getFeedbackWithPatients()
     {
         $feedback = PatientFeedback::orderBy('created_at', 'desc')->get();
         $patients = Patient::all();
+        if (auth()->user()->xss_stored_on) {
```
c-a-ray, line 35: > Suggestion

The duplicate returns are unnecessary. You could improve the flow a bit like this
```php
public function getFeedbackWithPatients()
{
    $feedback = PatientFeedback::orderBy('created_at', 'desc')->get();
    if (!auth()->user()->xss_stored_on) {
        $feedback = $this->sanitizeStoredFeedback($feedback);
    }
    return ['feedback' => $feedback, 'patients' => Patient::all()];
}
```

c-a-ray (commented):
Good stuff. Seems like everything is working correctly. The only thing that might be a requirements issue is that the stored feedback is always sanitized in `searchFeedback`, regardless of whether `xss_stored_on` is true.

All of the other comments are suggestions to improve code quality and are completely optional. I included "Suggestion" at the top of all of those.

app/Services/FeedbackService.php
Swangeon: I didn't want to do that here as I feel if there is JS in the stored feedback when you are trying to do the reflective XSS then I wanted to sanitize the stored feedback so that it doesn't get in the way of the reflective XSS attack. If we're still unsure about this then I can do the sanitization if `xss_stored_on` is false.

c-a-ray: Let's talk about this at our meeting tomorrow. I do kind of think we should sanitize unless `xss_stored_on` is false, because the user could have both stored and reflected toggles on, which seems to me like it should mean that both reflected and stored payloads would be executed.

c-a-ray (approved):
Nice work. Approved, but let's talk tomorrow about whether to sanitize stored feedback when injecting reflected payloads. I think we probably should not sanitize the stored feedback if both `xss_stored_on` and `xss_reflected_on` are true.

PR #102, opened 2025-03-06
Add Dashboard Content

Swangeon:
Add content to the home page to describe the application and its purpose.
Change the messages for the vulnerability toggles page along with changing the `File Upload` to `Unrestricted File Upload`.


resources/views/dashboard.blade.php
```diff
@@ -9,9 +9,23 @@
...
+                        <h1 class="text-2xl font-bold text-gray-900 dark:text-gray-100 content-center">Welcome to the BeaverHealth Vulnerable Web Application Dashboard!</h1>
+                    </div>
+                    <br>
+                    <p>BeaverHealth Vulnerable Web Application Dashboard is a PHP Laravel/MySQL web application that is meant to be vulnerable to various common web application security vulnerabilities found in the OWASP Top 10. The aim of the BeaverHealth Vulnerable Web Application Dashboard is to provide a platform for users to learn about and practice various web application security vulnerabilities in a safe and controlled environment with ample documentation to help the user learn the common web application security vulnerabilities. All of the vulnerabilities within the applicaiton that are implemented have in-depth documentation written about what they are, how to exploit them, and how to fix them where all of this information can be found on the wiki page of the project <a href="https://github.com/BeaverHealth-Vulnerable-Web-App/BeaverHealth-Vulnerable-Web-App/wiki" class="font-medium text-blue-600 dark:text-blue-500 hover:underline">here</a>.</p>
```
c-a-ray, line 17: Since this is on the homepage I think we want to get this as polished and professional as possible. Couple things:
- It's a little _too_ close to DVWA's message. Parts of it are almost exactly the same. We don't want to come off as a copy.
- The term "web application security vulnerabilities" is overused (3x in a short paragraph)
- A few parts are phrased a little awkwardly (e.g., "meant to be vulnerable," "help the user," "All of the vulnerabilities within the application that are implemented...," "...and how to fix them where all of this information can be found on the wiki...")
- "Application" is misspelled  ("applicaiton")
- It's a little verbose/wordy. We could get the same information across with a bit more brevity.

I'd suggest something more like this:

>Welcome to BeaverHealth Vulnerable Web Application
>
>BeaverHealth is a deliberately vulnerable PHP Laravel/MySQL web application designed to help users learn about common web security flaws in a hands-on environment. It provides a safe, controlled platform to explore a subset of vulnerabilities outlined in the OWASP Top 10, complete with documentation on how to identify, exploit, and remediate each issue.
>
>Our goal is to help security researchers, developers, and students build a deeper understanding of web application security through practical exploration.

resources/views/dashboard.blade.php
```diff
@@ -9,9 +9,23 @@
...
+                    <br>
+                    <h2 class="text-xl font-bold">WARNING!</h2>
+                    <br>
+                    <p>This website is intended for educational purposes only. We are not responsible for any damage or loss that may occur due to the use of this website. <em>Do not publically expose this application to the internet on any machine you control, as it will be compromised.</em> It is recommend using a virtual machine (such as VirtualBox or VMWare) and setting up the virtual machine to NAT networking mode.</p>
```
c-a-ray, line 21: Couple things:
- Not a fan of the liability statement. "We are not responsible for any damage or loss..." feels unnecessary and overly broad. DVWA doesn't even have anything like that.
- It's perfectly fine to run this outside of a VM, so I would like to exclude that bit. We should just emphasize that it shouldn't be exposed to the internet.

How about something like this:

> This application is designed for educational and testing purposes only. It contains known security vulnerabilities and should never be exposed to the internet or deployed in a production environment. Use it in a controlled setting and ensure it is not accessible from external networks. Its purpose is clearly defined, and we do not take responsibility for any misuse.

resources/views/dashboard.blade.php
```diff
@@ -9,9 +9,23 @@
...
+                    <br>
+                    <h2 class="text-xl font-bold">Links</h2>
+                    <br>
+                    <p><a href="https://github.com/BeaverHealth-Vulnerable-Web-App/BeaverHealth-Vulnerable-Web-App" class="font-medium text-blue-600 dark:text-blue-500 hover:underline">GitHub</a></p>
+                    <p><a href="https://github.com/BeaverHealth-Vulnerable-Web-App/BeaverHealth-Vulnerable-Web-App/wiki" class="font-medium text-blue-600 dark:text-blue-500 hover:underline">Documentation</a></p>
```
c-a-ray, line 26: The same link is provided in the text above. We should either have it here or there, not both places.

resources/views/vuln_toggles/index.blade.php
```diff
@@ -20,33 +20,33 @@
                         ]">
                             @php
                             $vulnerabilities = [
-                                'sqli_on' => ['SQL Injection', 'Change Password'],
-                                'cmd_inject_on' => ['Command Injection', 'Request Medical Records'],
-                                'bac_on' => ['Broken Access Control', 'Admin, Patient Info, Add Records, and Request Records'],
-                                'file_upload_on' => ['File Upload', 'Add Medical Records'],
-                                'xss_stored_on' => ['Stored Cross Site Scripting', 'Patient Feedback'],
-                                'xss_reflected_on' => ['Reflected Cross Site Scripting', 'Patient Feedback'],
+                            'sqli_on' => ['SQL Injection', 'Change Password'],
```
c-a-ray, line 23: Least important comment in here, but why unindent this stuff? Just makes it harder to read. Pretty sure it's a best-practice/standard. If you look at the official documentation (https://laravel.com/docs/11.x/blade), they indent the stuff between the directives.

c-a-ray (changes requested):
Nice work. Apologize for being super particular here. Just want to make sure to get this right since it'll be the first thing users see when they log in.

resources/views/vuln_toggles/index.blade.php
Swangeon: Genuinely have no idea why my editor tried to "fix" the formatting here as I agree that nothing was wrong with it before.

resources/views/dashboard.blade.php
Swangeon: Agreed with some of what you're saying here in terms of the wording but I would mention that while yes it **can** run fine outside of a VM, it **shouldn't** be ran outside of a VM because running it on your host machine is directly setting it to be exposed to the internet in some way (especially with how the IP that the container binds to is `0.0.0.0`).

c-a-ray: Still don't think it makes sense to include it on the homepage. When users read this message, they will already be running the app. A warning like that would make more sense in the README or something (i.e. something they see before running the app).

Also
> running it on your host machine is directly setting it to be exposed to the internet in some way

is not accurate. The container does bind to `0.0.0.0:9991`, but that doesn't inherently expose it to the internet. In most cases you would need to take deliberate steps (like port forwarding) to expose it publicly.

resources/views/dashboard.blade.php
```diff
@@ -9,9 +9,23 @@
...
             <div class="bg-white dark:bg-gray-800 overflow-hidden shadow-sm sm:rounded-lg">
                 <div class="p-6 text-gray-900 dark:text-gray-100">
-                    {{ __("You're logged in!") }}
+                    <div class="flex justify-center">
+                        <h1 class="text-2xl font-bold text-gray-900 dark:text-gray-100 content-center">Welcome to the BeaverHealth Vulnerable Web Application Dashboard</h1>
+                    </div>
+                    <br>
+                    <p>The BeaverHealth Vulnerable Web Application is a deliberately vulnerable PHP Laravel/MySQL web application designed to educate users about common web security vulnerabilities from the OWASP Top 10. It provides a safe, controlled environment with comprehensive documentation on each vulnerability, including what they are, how to exploit them, and how to fix them to help security researchers, developers, and students learn through hands-on experience.</p>
+                    <br>
+                    <h2 class="text-xl font-bold">WARNING!</h2>
```
c-a-ray, line 18: The "WARNING!" probably should be smaller than the welcome message. Might also want to change it to "Disclaimer" (see comment below).

resources/views/dashboard.blade.php
```diff
@@ -9,9 +9,23 @@
...
+                    <br>
+                    <p>This application is intended for educational and testing purposes only, and we do not take responsibility for any misuse. <em>Do not publically expose this application to the internet on any machine you control, as it contains vulnerabilities that are not intended to be used in a production environment or exposed to the internet.</em> It is recommend using a virtual machine (such as VirtualBox or VMWare) and setting up the virtual machine to NAT networking mode for the best security of your host machine when running the application.</p>
```
c-a-ray, line 20: Why don't we just have a disclaimer about intended use on the hompage, and put all of the stuff about not exposing it to the internet in the README? Like I said in the other message, users will already be running the app when they see this, so that information isn't very helpful here.

Also, "publicly" is misspelled (as "publically")

resources/js/vulnerability_toggles.js
```diff
@@ -24,12 +24,12 @@ document.addEventListener('DOMContentLoaded', function() {
 
                 const data = await response.json();
                 if (data.success) {
-                    showStatusMessage(`Successfully updated ${vulnName} toggle to ${isChecked ? 'on' : 'off'}`, 'success');
+                    showStatusMessage(`${isChecked ? 'Enabled' : 'Disabled'} ${vulnName}`, 'success');
```
c-a-ray, line 27: Nice, I think this makes the messages a lot clearer

c-a-ray (changes requested):
Nice work, almost there. Just a couple things

README.md
```diff
@@ -4,11 +4,17 @@
...
 2. Clone the repo 🌀
 3. Navigate to the project root 🫚  
   `cd ./BeaverHealth-Vulnerable-Web-App`
 4. Run the setup script
   `./setup.sh --fresh`
 5. In a web browser, navigate to `localhost:9991` 🗺️  
+
+## Disclaimer
+
+This application is intended for educational and testing purposes only, and we do not take responsibility for any misuse. Do not publicly expose this application to the internet on any machine you control, as it contains vulnerabilities that are not intended to be used in a production environment or exposed to the internet. It is recommend using a virtual machine (such as VirtualBox or VMWare) and setting up the virtual machine to NAT networking mode for the best security of your host machine when running the application.
```
c-a-ray, line 20: Nice, definitely makes more sense here than on the homepage. Just a couple things.

First, it's labelled "disclaimer," but everything after the first sentence is more of a usage recommendation than a disclaimer. The purpose of a disclaimer is typically only to limit liability and clarify responsibility. I'd recommend creating another header, "Warning," and moving everything after the first sentence there.

A couple of grammatical issues:
- "It is recommend using a virtual machine" should be "It is recommended to use a virtual machine" or "Using a virtual machine is recommended"
- There's a parallelism inconsistency in that same sentence, even after fixing the above issue: "to use a virtual machine ... and setting up the virtual machine..."
  - Corrected: "It is recommended to use a virtual machine ... and to set it up in NAT mode"

Some style issues:
- In this sentence: "Do not publicly expose this application to the internet on any machine you control, as it contains vulnerabilities that are not intended to be used in a production environment or exposed to the internet."
  - "on any machine you control" is unnecessary because it is clearly implied. It would be impossible for the reader to expose the app to the internet on any machine they _don't_ control.
  - "Do not publicly expose this application to the internet ... as it contains vulnerabilities that are not intended ... to be exposed to the internet." is repetitive and doesn't really communicate the real issue (that it could result in being compromised).
  - "publicly ... to the internet" is kind of redundant. They basically mean the same thing in this context.
    - Clearer version of that sentence: "Do not expose this application to the internet, as it contains vulnerabilities that could be exploited, potentially compromising your system."

Regarding the virtual machine stuff:
- I don't think it's necessary to give examples of virtual machine software. No reason to give free advertising.
- I still think we should make it clear that it's not always necessary to use a virtual machine. It's really only needed if you aren't confident about your network security settings.

How about this?
### Disclaimer
This application is intended for educational and testing purposes only, and we do not take responsibility for any misuse.

### Warning
Do not expose this application to the internet, as it contains vulnerabilities that could be exploited, potentially compromising your system. If you are not fully confident in your network security settings, consider running it in a virtual machine with NAT mode to better isolate it from your host system."

resources/views/dashboard.blade.php
```diff
@@ -9,9 +9,19 @@
...
             <div class="bg-white dark:bg-gray-800 overflow-hidden shadow-sm sm:rounded-lg">
                 <div class="p-6 text-gray-900 dark:text-gray-100">
-                    {{ __("You're logged in!") }}
+                    <div class="flex justify-center">
+                        <h1 class="text-2xl font-bold text-gray-900 dark:text-gray-100 content-center">Welcome to the BeaverHealth Vulnerable Web Application</h1>
+                    </div>
+                    <br>
+                    <p>BeaverHealth is a deliberately vulnerable PHP Laravel/MySQL web application designed to educate users about a subset of common web security vulnerabilities from the OWASP Top 10. It provides a safe, controlled environment with comprehensive documentation on what each vulnerability is, how to exploit them, and how to fix them. Our goal with this project is to help security researchers, developers, and students build a deeper understanding of web application security through hands-on experience.</p>
+                    <br>
+                    <h2 class="text-xl font-bold">Links</h2>
```
c-a-ray, line 18: "Links" is bigger/bolder than the header. At very least, we should make it smaller/lighter. Though i don't think it's necessary to have there at all, as it's already pretty clear they are links. If you want to make it more clear, I'd suggest making them underlined or something.

Also, the links themselves are bigger than the header and body text. They should at least be smaller than the header, and at least the same size as the body text.

resources/views/dashboard.blade.php
```diff
@@ -9,9 +9,19 @@
         <div class="max-w-7xl mx-auto sm:px-6 lg:px-8">
             <div class="bg-white dark:bg-gray-800 overflow-hidden shadow-sm sm:rounded-lg">
                 <div class="p-6 text-gray-900 dark:text-gray-100">
-                    {{ __("You're logged in!") }}
+                    <div class="flex justify-center">
+                        <h1 class="text-2xl font-bold text-gray-900 dark:text-gray-100 content-center">Welcome to the BeaverHealth Vulnerable Web Application</h1>
```
c-a-ray, line 13: The header is the same size as the body text (actually looks slightly lighter too). Should be a little bigger/bolder

c-a-ray (changes requested):
Getting closer. Still a few grammatical errors and other small issues.

resources/views/dashboard.blade.php
Swangeon: Don't quite get what you mean here?
image
I edited the style so that they are right next to each other, but the header is bigger than body text.

resources/views/dashboard.blade.php
Swangeon: again, don't quite get what you mean here:
image
Links is smaller than the header and all the body text, including the links, are the same.

c-a-ray: Not sure what was going on, my bad. For some reason the styling wasn't getting applied correctly when I was looking at it yesterday. Looks fine now.

I still think we could do without the "Links" header, but I won't fight you on that.

resources/views/dashboard.blade.php
c-a-ray: Yeah, my bad. Something must have been going wrong on my end.

resources/views/dashboard.blade.php
```diff
@@ -9,9 +9,19 @@
...
             <div class="bg-white dark:bg-gray-800 overflow-hidden shadow-sm sm:rounded-lg">
                 <div class="p-6 text-gray-900 dark:text-gray-100">
-                    {{ __("You're logged in!") }}
+                    <div class="flex justify-center">
+                        <h1 class="text-2xl font-bold text-gray-900 dark:text-gray-100 content-center underline">Welcome to the BeaverHealth Vulnerable Web Application</h1>
+                    </div>
+                    <br>
+                    <p>BeaverHealth is a deliberately vulnerable PHP Laravel/MySQL web application designed to educate users about a subset of common web security vulnerabilities from the OWASP Top 10. It provides a safe, controlled environment with comprehensive documentation on what each vulnerability is, how to exploit them, and how to fix them. Our goal with this project is to help security researchers, developers, and students build a deeper understanding of web application security through hands-on experience.</p>
+                    <br>
+                    <h2 class="text-xl font-bold underline">Links</h2>
```
c-a-ray, line 18: Still think it looks cleaner without the "Links" header, but I won't fight you on that if you really want to keep it.

c-a-ray (changes requested):
So close! 

There's just still one grammatical issue in the homepage text. Also realized there's a small preexisting issue in the README. Figured you wouldn't mind fixing it real quick since you're already in there.

resources/views/dashboard.blade.php
```diff
@@ -9,9 +9,19 @@
         <div class="max-w-7xl mx-auto sm:px-6 lg:px-8">
             <div class="bg-white dark:bg-gray-800 overflow-hidden shadow-sm sm:rounded-lg">
                 <div class="p-6 text-gray-900 dark:text-gray-100">
-                    {{ __("You're logged in!") }}
+                    <div class="flex justify-center">
+                        <h1 class="text-2xl font-bold text-gray-900 dark:text-gray-100 content-center underline">Welcome to the BeaverHealth Vulnerable Web Application</h1>
+                    </div>
+                    <br>
+                    <p>BeaverHealth is a deliberately vulnerable PHP Laravel/MySQL web application designed to educate users about a subset of common web security vulnerabilities from the OWASP Top 10. It provides a safe, controlled environment with comprehensive documentation on what each vulnerability is, how to exploit them, and how to fix them. Our goal with this project is to help security researchers, developers, and students build a deeper understanding of web application security through hands-on experience.</p>
```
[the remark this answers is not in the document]
Swangeon: Will do, this is the new paragraph and to Grammarly it's good now.
image

c-a-ray (approved):
Nice work.

Approving as-is. Let's have a group conversation about that "Links" header at some point. It just looks tacky. But we've been round and round on this PR so I don't want to hold it up any longer.

PR #104, opened 2025-03-09
Insecure Change Password

c-a-ray:
This PR makes the Change Password page vulnerable to SQL injection when the `sqli_on` toggle is enabled. To make this possible, I had to add a "username" field to the form and use a raw `UPDATE` statement in `ChangePasswordService` that references the value entered in that field. Due to the limitations of the Change Password page, SQL injection can only be used to update the current user's password without needing to enter the correct current password.

To test the functionality, you can enter something like this:
Username Confirmation: `' OR '1'='1`
Current Password: notmyrealpassword
New Password: newpassword
New Password Confirmation: newpassword

Through doing this I've come to the conclusion that the Change Password page is not an ideal place to demonstrate SQL injection:
- The page relies entirely on an `UPDATE` statement to modify the current user's password, which does not return data.
- Because we need to use `$user->user_id` in the query to maintain the page's functionality, the user can only update their own record (or all records--see below).
- Initially, I wrote the query like this
```sql
UPDATE user SET password = '$hashedNewPassword'
WHERE user_id = {$user->user_id}
AND username = '$usernameConfirmation'
```
But this results in ALL users' password being updated when a boolean expression is injected (e.g. `UPDATE user SET password = 'hash' WHERE user_id = 1 AND USERNAME = '' OR '1'='1'` evaluates to `UPDATE user SET password = 'hash' WHERE TRUE`).

I do not think we want to do that. To make it so that doesn't happen, I had to use parenthesis to change the precedence:
```sql
UPDATE user SET password = '$hashedNewPassword'
WHERE user_id = {$user->user_id}
AND (username = '$usernameConfirmation')
```
But this effectively makes it impossible to stack queries, which limits what is possible.

If anyone has any ideas for how to make this work, I'm all ears.

I'm basically now of the opinion that we should figure out somewhere else to demonstrate SQL injection. Doing it on a page that uses a query to grab data from the database would probably be the best option. Modifying the Patient Info page to contain an input field may be the easiest way to do this.

For now, though, the implementation here gets the job done. It allows users to change the authenticated user's password without knowing what the current password is, which is a genuine vulnerability. It is just more limited than I was hoping. It at least gives us something to demo for alpha.

### DIFF
--- a/app/Http/Controllers/PasswordController.php
+++ b/app/Http/Controllers/PasswordController.php
@@ -27,10 +27,10 @@ class PasswordController extends Controller
     public function changePassword(ChangePasswordRequest $request): RedirectResponse
     {
         $user = Auth::user();
-
         try {
-            $success = $this->changePasswordService->updatePassword($user, $request);
-            if ($success) {
+            $result = $this->changePasswordService->updatePassword($user, $request);
+
+            if ($result['success']) {
                 session()->flash('change-password-status', [
                     'type' => 'success',
                     'message' => 'Password updated',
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/PasswordController.php
```diff
@@ -27,17 +28,26 @@ public function changePassword(ChangePasswordRequest $request): RedirectResponse
...
+                    'message' => 'Password updated',
+                ]);
+                return redirect()->route('profile.change-password');
+            }
+
+            return redirect()->route('profile.change-password')
+                ->withErrors(['current_password' => 'Current password is incorrect']);
+        } catch (QueryException $e) {
+            if ($user->sqli_on) {
+                return redirect()->route('profile.change-password')
```
c-a-ray, line 45: Used to display SQL errors to the user (as an educational tool)

PR #105, opened 2025-03-09
Secure Versions of Add & Request Medical Records Pages

BrynnMartina:


resources/views/records/request.blade.php
BrynnMartina, line 1: Status messages for the request page weren't updated to use the universal status message formatting

resources/views/records/add.blade.php
```diff
@@ -39,11 +44,28 @@ class="w-full px-6 py-4 bg-gray-500 text-white rounded shadow-md hover:bg-gray-6
...
     </div>
+
     <!-- Status message -->
     @if (session('records-status'))
         <x-status-message
             :message="session('records-status')['message']"
             :type="session('records-status')['type']"
         />
     @endif
+
```
BrynnMartina, line 55: Code to prevent laravel crash page from popping up if file size exceeds 100MB

resources/views/records/add.blade.php
```diff
@@ -39,11 +44,28 @@ class="w-full px-6 py-4 bg-gray-500 text-white rounded shadow-md hover:bg-gray-6
...
+            if (file && file.size > maxSize) {
+                document.getElementById('file-size-error').innerText = 'File size exceeds the maximum limit of 100MB.';
+                document.getElementById('file-size-error').style.display = 'block';
+                input.value = '';
+            } else {
+                document.getElementById('file-size-error').style.display = 'none';
+            }
+        }
+    </script>
```
BrynnMartina, line 70: Code to prevent laravel crash page from popping up when submitting a file over 100MB

resources/views/records/add.blade.php
```diff
@@ -39,11 +44,28 @@ class="w-full px-6 py-4 bg-gray-500 text-white rounded shadow-md hover:bg-gray-6
...
     @if (session('records-status'))
         <x-status-message
             :message="session('records-status')['message']"
             :type="session('records-status')['type']"
         />
     @endif
+
+    <!-- JavaScript for File Size Validation -->
+    <script>
+        function validateFileSize(input) {
```
c-a-ray, line 58: > Suggestion(s)

We have all of our other JavaScript stored in `resources/js/`. Probably best to put this in something like `resources/js/records.js` for consistency. 

Also, if you put it in there, you could add a constant (maybe in `utils.js`) so you could replace the magic number (`100 * 1024 * 1024`).

resources/views/records/add.blade.php
```diff
@@ -39,11 +44,28 @@ class="w-full px-6 py-4 bg-gray-500 text-white rounded shadow-md hover:bg-gray-6
...
+            const maxSize = 100 * 1024 * 1024;
+            const file = input.files[0];
+
+            if (file && file.size > maxSize) {
+                document.getElementById('file-size-error').innerText = 'File size exceeds the maximum limit of 100MB.';
```
c-a-ray, line 63: > Important

Should probably say `100 MiB` for clarity. A lot of software would display this as `104.8 MB`.

resources/views/records/request.blade.php
```diff
@@ -26,15 +26,17 @@ class="w-full p-2 border border-gray-300 dark:border-gray-600 rounded bg-white d
                         </div>
 
                         @if (session('success'))
-                            <div class="mt-4 p-2 text-green-800 bg-green-100 border border-green-500 rounded">
-                                {{ session('success') }}
-                            </div>
+                            <x-status-message
```
c-a-ray, line 29: > Sort of important

To stay consistent with how we do this on other pages and prevent session key conflicts, probably best to use a more specific session key and store the message and type in an associative array so you could have a single `<x-status-message>` (see `profile/change-password.blade.php`, `feedback/index.php`, `records/add.blade.php`, and `layouts/app.blade.php`). E.g.,

```php
    @if(session('records-request-status'))
        <x-status-message
            :message="session('records-request-status')['message']"
            :type="session('records-request-status')['type']"
        />
    @endif
```

c-a-ray (changes requested):
Great work. For the most part everything looks really good. 

All of my comments are pretty minor. At the top of each of them I included
> Suggestion

or

> Important

Feel free to ignore the ones marked "suggestion" if you disagree. Would like to at least talk about the ones marked "important" though.

The important things are:
- I think we probably want to allow file types that would more realistically be used for medical records.
- Probably want to check the MIME type as well as the file extension, to prevent extension spoofing.
- Probably should format the file size error messages to use "MiB" instead of "MB" for clarity.

The others are mostly just little code style things.

resources/js/records.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Validates the file size when a file is selected
+ *
+ * @param {HTMLInputElement} input
+ */
+function validateFileSize(input) {
+    const file = input.files[0];
+
+    if (file && file.size > MAX_FILE_SIZE_BYTES) {
```
c-a-ray, line 11: Looks like this needs to be a `>=` instead of `>`. When I use a file that's exactly 100 MiB, it isn't caught by the client side logic and causes the server error.

```suggestion
    if (file && file.size >= MAX_FILE_SIZE_BYTES) {
```

c-a-ray (changes requested):
Looks great! Just found one issue with loading files that are exactly 100 MiB. Simple fix

c-a-ray (approved):
Nice work!

PR #107, opened 2025-03-11
Fix File Display

BrynnMartina:
Gets rid of dots displaying when showcasing a user's files on the request records page (caused by the new scandir)


c-a-ray (approved):
lgtm

PR #120, opened 2025-04-13
Hardening, logging, and config cleanup

c-a-ray:
This started as the demo deployment branch but evolved into a full application hardening and infrastructure cleanup effort. There are quite a few changes, but pretty much everything is related to security/logging/configuration, so I figured it made the most sense to put it all in this PR.

Configuration/environment cleanup:
- Replaces the committed `.env` file with a proper `.env.example`. This is something I've been wanting to do for a while. I think it's usually considered bad practice to have a real `.env` file in source control.
- Refactors the `docker-compose` files into a base config (`docker-compose.yaml`) and local overrides (`docker-compose.dev.yaml`) useful for development environments. Mostly just simplifies the configs and removes duplication.
- Updates CI to deal with `.env` not being in source control (it creates a `.env` file from the `.env.example` and gets the `APP_KEY` from a GitHub secret).
- Cleans up all config files for clarity and removes unused or misleading environment variables
- Updates the `setup.sh` script to create a valid `.env` and generate a secure `APP_KEY` if missing

Security and logging improvements:
- Adds a `UserActivityLogger` service, used for logging across the app.
- Adds logging for all route accesses (registered and unregistered), as well as point-of-action user activity (most of this is in the service classes--I put the logging there instead of in the controllers so more useful context could be included in the logs).
- Adds `AjaxMiddleware` to restrict `/sidebar/refresh/` to AJAX requests only (currently there's a bug where you can navigate to `/sidebar/refresh`).
- Adds rate limiting for login page accesses. This was in response to a large number of login page accesses from automated scanners when the app was deployed on GCP.
- Adds `TrustProxiesMiddleware` to ensure correct client IP resolution behind proxies (used only in `prod` environment). Basically, when the app was behind a load balancer, the client IP always showed up as the load balancer's IP, which breaks rate limiting (because rate limiting is keyed on IP) and makes logging less meaningful. This middleware adds trusted proxies and sets some headers to allow the real client IP to be forwarded.

Other changes:
- Adds a dedicated `patient_records` disk to store uploaded files (currently we use `app/public`  with "public" visibility, which doesn't really make sense, since the files aren't served). There was also a weird issue where the `patient_records` directory was created in both `app/public` and `app/private`. This fixes that, too. This change doesn't impact insecure file upload or command injection vulnerabilities.
- Adds a new App Reset feature accessible via the web interface
- Refactors service classes in a few places for cleaner logging. I refactored a few other things while I was doing this, too. Nothing major--just obvious/easy improvements.
- Updates user seeding logic. When environment is `local`, only a `dev` user is created (no more `admin` user--I don't think an "admin" user really makes sense). I kept the seeder logic for production, as well, and added a switch to make it only execute in the `prod` environment.
- Adds consistent PHPDoc comments and type hints. I didn't do this for every single file. Only the ones that I created in the past. Figured I'd let other team members add PHPDoc comments to the controllers/services they created.
- Changes the `file_upload_on` field to `file_upload_vuln_on`. This is the most unrelated change here. I did it because `file_upload_on` is a super confusing field name (that's my bad, since I created it in the first place).


database/seeders/UserSeeder.php
```diff
@@ -7,27 +7,78 @@
...
-        // Create 10 fake regular users
-        User::factory()->count(10)->create();
+    /**
+     * Create production users with standard permissions.
+     *
+     * @return void
+     */
+    private function createProdUsers(): void
+    {
+        $this->createUser('gokul', 'eN%f0G6M!xGiX@^K7$');
```
c-a-ray, line 46: We'll definitely want to change these passwords when we go to deploy again, but I figured it wouldn't hurt to leave this here until then.

.env.example
```diff
@@ -1,44 +1,48 @@
-APP_NAME=Laravel
+APP_NAME=BeaverHealth
 APP_ENV=local
-APP_KEY=base64:yw6BR9k5FinsLSVQXaeKFNeWkGTU5sR+jwci/WvAzN0=
+APP_KEY=!!!REPLACE!!!
```
c-a-ray, line 3: The setup script handles turning this into `.env` and replacing the `APP_KEY` with a real one.

.env.example
```diff
@@ -1,44 +1,48 @@
...
+LOG_CHANNEL=single
+LOG_DEPRECATIONS_CHANNEL=deprecation
 LOG_LEVEL=debug
 
 DB_CONNECTION=mysql
 DB_HOST=db
 DB_PORT=3306
 DB_DATABASE=vuln_db
-DB_USERNAME=admin
+DB_USERNAME=dev
```
c-a-ray, line 24: I think a "dev" user makes more sense than an "admin" user. Actual end-users will use the registration feature to create their own user account. The dev user is just for developers.

.github/workflows/ci.yaml
```diff
@@ -18,16 +18,23 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
 
+      - name: Prepare .env
+        run: |
+          cp .env.example .env
+          sed -i 's/^APP_ENV=.*/APP_ENV=ci/' .env
+          sed -i 's/^APP_DEBUG=.*/APP_DEBUG=false/' .env
+          sed -i "s|^APP_KEY=.*|APP_KEY=${{ secrets.CI_APP_KEY }}|" .env
```
c-a-ray, line 26: Sets up a `.env` file appropriate for CI, and gets the `APP_KEY` from a GitHub secret.

.gitignore
```diff
@@ -1,23 +1,11 @@
...
-/public/hot
 /public/storage
-/storage/*.key
-/storage/pail
+/node_modules
 /vendor
-.env
-.env.backup
-.env.production
-.phpactor.json
```
c-a-ray, line 12: I think these were just defaults. Deleted things that we don't actually have. Let me know if anybody actually has any of these things locally.

app/Http/Controllers/AdminController.php
```diff
@@ -3,30 +3,29 @@
...
-    private UserRoleService $roleService;
-
     /**
-     * Create a new controller instance.
+     * Create a new AdminController instance.
      *
      * @param UserRoleService $roleService The role service for handling permission operations
      */
-    public function __construct(UserRoleService $roleService)
+    public function __construct(private UserRoleService $roleService)
```
c-a-ray, line 19: I learned that you can inject services just by doing this. I think it's a lot cleaner than the way we were doing it before. I didn't update every file to do this (just the ones I was already messing with).

app/Http/Controllers/AdminController.php
```diff
@@ -3,30 +3,29 @@
...
     {
-        $this->roleService = $roleService;
     }
 
     /**
      * Display the admin panel with all users.
      *
-     * @return \Illuminate\View\View The admin view with users data
+     * @return View The admin view with users data
```
c-a-ray, line 26: To move toward consistency, I adopted this standard:
- Always provide a PHPDoc comment for methods
- Don't use the full qualified path for types
- Provide type hints for all parameters and return values

I didn't do this for all files, just the ones I created.

app/Http/Controllers/RegisteredUserController.php
```diff
@@ -45,7 +45,7 @@ public function register(Request $request): RedirectResponse
                 'load_records' => true,
                 'view_patient_info' => false,
                 'sqli_on' => false,
-                'file_upload_on' => false,
+                'file_upload_vuln_on' => false,
```
c-a-ray, line 48: As mentioned in the PR description, `file_upload_on` is just a super confusing field name. It makes it sound like it determines whether file uploads are enabled at all. This change makes it more clear that its the vulnerability that is enabled/disabled.

app/Http/Middleware/AjaxMiddleware.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+class AjaxMiddleware
```
c-a-ray, line 9: See the PR description for the reasoning behind this. Fixes a bug.

app/Http/Middleware/CheckPermissionMiddleware.php
```diff
@@ -5,33 +5,44 @@
...
-    * @param  \Illuminate\Http\Request  $request
-    * @param  \Closure  $next
-    * @return mixed
+    * @param Request $request The current HTTP request
+    * @param Closure $next    The next middleware handler
+    *
+    * @return Response A redirect response if access is denied, or the response from the next middleware/controller
     */
-    public function handle(Request $request, Closure $next)
+    public function __invoke(Request $request, Closure $next): Response
```
c-a-ray, line 32: Changed to using `__invoke()` so I could use the logger service.

app/Http/Middleware/LogUserActivityMiddleware.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+use App\Services\UserActivityLogger;
+use Symfony\Component\HttpFoundation\Response;
+
+class LogUserActivityMiddleware
```
c-a-ray, line 10: This is the thing that logs all route accesses.

app/Http/Middleware/LoginPageRateLimiter.php
```diff
@@ -0,0 +1,38 @@
...
+
+namespace App\Http\Middleware;
+
+use App\Services\UserActivityLogger;
+use Closure;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\RateLimiter;
+use Symfony\Component\HttpFoundation\Response;
+
+class LoginPageRateLimiter
```
c-a-ray, line 11: Added because the login page was getting hammered by automated scanners when the app was deployed on GCP.

app/Http/Middleware/TrustProxiesMiddleware.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Illuminate\Http\Middleware\TrustProxies;
+use Illuminate\Http\Request;
+
+class TrustProxiesMiddleware extends TrustProxies
```
c-a-ray, line 8: See PR description and the PHP Docs here for reasoning behind this.

app/Services/FeedbackService.php
```diff
@@ -35,18 +40,32 @@ public function getFeedbackWithPatients()
 
     public function storeFeedback(FeedbackCommentRequest $request)
     {
-        $xssStoredToggle = auth()->user()->xss_stored_on;
+        $user = auth()->user();
+        $patientId = $request->input('patient_id');
```
c-a-ray, line 44: Moved these up here so I could log them

app/Services/FeedbackService.php
```diff
@@ -35,18 +40,32 @@ public function getFeedbackWithPatients()
...
+            'xss_enabled' => $user->xss_stored_on
+        ]);
 
         PatientFeedback::create([
-            'patient_id' => $request->input('patient_id'),
-            'feedback' => $this->processInput($request->input('feedback'), $xssStoredToggle),
-            'is_vulnerable' => $xssStoredToggle ? true : false
+            'patient_id' => $patientId,
+            'feedback' => $this->processInput($feedback, $user->xss_stored_on),
+            'is_vulnerable' => $user->xss_stored_on
```
c-a-ray, line 56: `xss_stored_on` is already a bool, so no need for the ternary

app/Services/PatientRecordService.php
```diff
@@ -6,23 +6,116 @@
...
 use App\Presenters\FilePresenter;
+use App\Services\UserActivityLogger;
 
 class PatientRecordService
 {
-    protected $filePresenter;
-    private const MAX_FILE_SIZE_BYTES = 5 * 1024 * 1024;
+    private const MAX_FILE_BYTES = 5 * 1024 * 1024;
+
+    private const ALLOWED_EXTENSIONS = [
```
c-a-ray, line 15: Moved these up here as constants so it's easier to update in the future if we want.

app/Services/PatientRecordService.php
```diff
@@ -6,23 +6,116 @@
...
+        'application/json',
+        'application/EDI-X12',
+        'application/xml',
+        'text/xml',
+        'application/pdf',
+        'text/plain'
+    ];
+
+    public function __construct(
+        private FilePresenter $filePresenter,
```
c-a-ray, line 31: Converted to the new service injection approach because I had to inject the logger

app/Services/PatientRecordService.php
```diff
@@ -6,23 +6,116 @@
...
+
+        return PatientFile::create([
+            'patient_id' => $patientId,
+            'filename'   => $filename,
+            'path'       => $filePath
+        ]);
     }
 
-    public function listFiles($directory, $keyword = '', $isInsecure = false)
+    private function listFiles($directory, $keyword = '', $isInsecure = false)
```
c-a-ray, line 118: Made this private and moved it down here because it's only used in this file

app/Services/PatientRecordService.php
```diff
@@ -6,23 +6,116 @@
...
+            throw new \Exception($error);
+        }
+
+        $filename = $file->getClientOriginalName();
+        $fileExtension = strtolower($file->getClientOriginalExtension());
+        $fileSize = $file->getSize();
+        $fileMimeType = $file->getMimeType();
+
+        if (!$user->file_upload_vuln_on) {
+            $error = $this->validateFile($fileExtension, $fileMimeType, $fileSize);
```
c-a-ray, line 84: Added this `validateFile` method to make logging easier. Bonus is that it makes this method a lot easier to read.

app/Services/PatientRecordService.php
```diff
@@ -53,81 +146,62 @@ public function listFiles($directory, $keyword = '', $isInsecure = false)
...
+            return 'Invalid file type. Allowed MIME types: ' . implode(', ', self::ALLOWED_MIME_TYPES) . '.';
+        }
+        if ($size > self::MAX_FILE_BYTES) {
+            return "File too large. Maximum size allowed is 5MiB.";
         }
+        return null;
     }
 
-    public function storeRecord($patientId, $file)
+    private function storeFile($patientId, $file)
```
c-a-ray, line 163: This used to be `protected`, but it's only needed in this file, so I made it `private`.

app/Services/PatientRecordService.php
```diff
@@ -53,81 +146,62 @@ public function listFiles($directory, $keyword = '', $isInsecure = false)
...
-            }
-
-            if ($fileSize > self::MAX_FILE_SIZE_BYTES) {
-                throw new \Exception('File too large. Maximum size allowed is 5MiB.');
-            }
-        }
-
-        $filePath = $this->storeFile($patient->patient_id, $file);
+        $fileName = $file->getClientOriginalName();
+        return $file->storeAs($directory, $fileName, 'patient_records');
```
c-a-ray, line 169: Files now stored on a dedicated disk at `app/patient_records`

app/Services/UserActivityLogger.php
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @return void
+     */
+    private function log(string $message, array $context = [], string $level = 'info'): void
+    {
+        $user = Auth::user();
+        $request = request();
+
+        $baseContext = [
+            'user_id'    => $user?->user_id,
```
c-a-ray, line 51: Nullable because this is also used when the user isn't authenticated

app/Services/UserRoleService.php
```diff
@@ -2,43 +2,86 @@
...
+
     /**
      * Check if the current user is authorized to update roles.
      *
      * @return bool Whether the current user is authorized to update roles.
      */
     public function authorize(): bool
     {
-        $currentUser = auth()->user();
+        $currentUser = Auth::user();
```
c-a-ray, line 27: We're inconsistent with `auth()->user()` vs `Auth::user()`. I prefer the `Auth` facade, but I'm ok with using `auth()->user()` if everyone prefers that. We should just be consistent. From what I understand, they are functionally equivalent. This just looks nicer IMO.

bootstrap/app.php
```diff
@@ -1,24 +1,48 @@
...
+use App\Http\Middleware\LogUserActivityMiddleware;
+use App\Http\Middleware\AjaxMiddleware;
+use App\Http\Middleware\LoginPageRateLimiter;
+use App\Services\UserActivityLogger;
 
 return Application::configure(basePath: dirname(__DIR__))
     ->withRouting(
         web: __DIR__ . '/../routes/web.php',
-        health: '/up',
+        health: '/health',
```
c-a-ray, line 18: Changed this because the default health check on GCP was to `/health`

bootstrap/app.php
```diff
@@ -1,24 +1,48 @@
...
                 'check.permission' => CheckPermissionMiddleware::class,
+                'ajax' => AjaxMiddleware::class,
+                'login.page.limit' => LoginPageRateLimiter::class,
             ]);
         }
     )
     ->withExceptions(
         function (Exceptions $exceptions) {
-            //
+            $exceptions->renderable(
```
c-a-ray, line 35: This stuff is for logging attempts to access unregistered routes. This was happening A LOT when the app was deployed on GCP.

composer.json
```diff
@@ -2,10 +2,7 @@
     "name": "laravel/laravel",
     "type": "project",
     "description": "The skeleton application for the Laravel framework.",
-    "keywords": [
-        "laravel",
-        "framework"
-    ],
+    "keywords": ["laravel", "framework"],
```
c-a-ray, line 5: Didn't actually change anything here. Just auto-formatted.

config/app.php
```diff
@@ -2,125 +2,36 @@
...
-    | Application Name
-    |--------------------------------------------------------------------------
-    |
-    | This value is the name of your application, which will be used when the
-    | framework needs to place the application's name in a notification or
-    | other UI elements where an application name needs to be displayed.
-    |
-    */
+    // Application name
+    'name' => env('APP_NAME'),
```
c-a-ray, line 6: For all of the `config/` files, deleted everything that we definitely don't use and added more streamlined comments (so it's easier to read the files). For things like `APP_ENV` that definitely need to be set, I purposefully did not provide a fallback.

database/seeders/UserSeeder.php
```diff
@@ -7,27 +7,82 @@
...
+    }
+
+    /**
+     * Create production users with standard permissions.
+     *
+     * @return void
+     */
+    private function createProdUsers(): void
+    {
+        $this->createUser('gokul', env('GOKUL_USER_PASSWORD'));
```
c-a-ray, line 50: We just need to make sure to add these environment variables to the `.env` when we go to deploy at the end of the term. Wanted to keep the logic for creating prod users but didn't want to have hardcoded plaintext passwords. Not that it really matters, just seems like a bad look.

docker-compose.yaml
```diff
@@ -0,0 +1,22 @@
+services:
```
c-a-ray, line 1: This minimal version of the config has everything the app needs to run. `docker-compose.dev.yaml` can be stacked on top of it for things that make sense in a dev environment.

setup.sh
```diff
@@ -115,6 +120,40 @@ determine_actions() {
...
+  else
+    echo -e "${GREEN}Found .env"
+  fi
+}
+
+generate_app_key() {
+  if [[ "$GENERATE_APP_KEY" = true ]]; then
+    echo -e "${CYAN}Generating APP_KEY...${NO_COLOR}"
+
+    key="base64:$(head -c 32 /dev/urandom | base64)"
```
c-a-ray, line 141: I spent hours trying to get this to work by using `artisan key:generate`, but it was really difficult because the key needs to be there for the dependencies/build steps to work, and you can't use `artisan` until those steps are complete, so it was a chicken-and-egg nightmare.

Then I realized that `artisan key:generate` just returns a random 32-bit base64 string, and there's no reason we can't just generate the key some other way. Every Linux system will have `dev/urandom`, so I went with that.

setup.sh
```diff
@@ -181,7 +220,7 @@ wait_for_database() {
 
 setup_database() {
   if [[ "$MIGRATE_DB" = true ]]; then
-    if [[ "$FRESH_DEPLOYMENT" != true ]] && [[ "$WIPE_DB" = true ]]; then
+    if [[ "$WIPE_DB" = true ]]; then
```
c-a-ray, line 223: Realized we actually want to wipe the database when using `--fresh`. It doesn't hurt anything if there's nothing in the database to wipe, and it makes it so `--fresh` still works if there's been a schema change.

PR #124, opened 2025-04-13
Prod hardening

c-a-ray:
This PR adds some hardening measures that are mostly relevant to a prod environment.

- Adds a new `AjaxMiddleware` to fix a bug with the `/sidebar/refresh` route. The route is used to dynamically update the sidebar elements, but I found that you can force-browse to the route. This makes it so that can't happen.
- Adds a `LoginPageRateLimiter` middleware that rate limits login page accesses. When the app was deployed on GCP, the login page was getting hammered by automated scanners. This helps to mitigate that issue.
- Adds a `TrustProxiesMiddleware`. When the app was deployed behind a load balancer, the client IP was always showing up as the IP of the load balancer. This was a problem because it broke rate limiting and made logging less meaningful. This middleware adds trusted proxies and HTTP headers that make it possible to access the true client IP. It only adds trusted proxies when environment is `prod`, though. When not in a production environment, it trusts no proxies.
- Forces HTTPS when environment is `prod`
- Updates the health check route to `/health`, which was the default for GCP.
- Adds some logic to the `UserSeeder` to determine which users to create based on the environment. In `prod`, it creates users for devs and graders (passwords set via environment variables). In `local`, it creates only a `dev` user (password also set via an environment variable). Note that this moves away from the `admin` user. I think having a `dev` user makes a bit more sense, since that's really what it's used for.


BrynnMartina:
-Ajaxmiddleware bug seems to be fixed on my end.
-The rate limiter works after the 11th attempt. Error message does look sparse. If there's ever time, it would be nice for this to be some sort of pop up on the login page.
Screenshot from 2025-04-23 01-16-07
-Trust proxies middleware works as expected.
-localhost:9991/health shows the info.
-Seeder logic is sound. Changing to dev also makes sense.

Overall, everything works on my end! Just a potential UI mention.

BrynnMartina (approved):

PR #125, opened 2025-04-13
App reset feature

c-a-ray:
This PR adds the app reset feature.

Clicking the sidebar link takes you to a confirmation page. Confirming causes the app to delete all patient files, wipe and recreate the database, and log out the user.


app/Http/Controllers/AppResetController.php
```diff
@@ -0,0 +1,59 @@
...
+use App\Services\UserActivityLogger;
+use Illuminate\Support\Facades\Artisan;
+use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\File;
+use Illuminate\View\View;
+use Illuminate\Http\RedirectResponse;
+
+class AppResetController extends Controller
+{
+    /**
```
alherrera-cs, line 14: Quick question, should this be admin‑only or available to any signed‑in user? I can see pros and cons for both, curious what you are thinking.

resources/views/components/confirm-reset.blade.php
```diff
@@ -0,0 +1,43 @@
...
+
+                    <div class="flex justify-between">
+                        <a href="{{ route('dashboard') }}" class="bg-gray-500 hover:bg-gray-700 text-white font-bold py-2 px-4 rounded">
+                            Cancel
+                        </a>
+
+                        <form method="POST" action="{{ route('app.reset') }}">
+                            @csrf
+                            <button type="submit" class="bg-red-600 hover:bg-red-800 text-white font-bold py-2 px-4 rounded">
+                                Yes, Reset Everything
```
alherrera-cs, line 35: Maybe once the user clicks this, we should disable the button to prevent accidental repeat

alherrera-cs (commented):
Thanks for the work on this! Everything looks great. My only question is should we lock this down to admins only or leave it open for all signed‑in users?

app/Http/Controllers/AppResetController.php
c-a-ray: I think it should be available to every user.

resources/views/components/confirm-reset.blade.php
c-a-ray: The user gets logged out when this button is clicked, so it's not really possible to repeat click it

Swangeon (approved):
Looks all good to me here. having that will be a great advantage to the user if the application breaks on anyway on the backend

BrynnMartina (approved):

PR #126, opened 2025-04-15
SQL Injection on Patient information page

alherrera-cs:
This pull request adds an SQL injection vulnerable search field on the Patient Information page. When a user with sqli_on enabled searches, the input is directly inserted into the SQL query (e.g. entering:
' OR '1'='1  will cause the query to return all patient records. When sqli_on is off, a safe parameterized query is used. A try-catch block handles malformed payloads and displays error message. 

Testing Instructions:

   - Log in as admin with sqli_on enabled.

   - In the search field, enter: ' OR '1'='1 to see all patient records.


app/Services/PatientInfoService.php
```diff
@@ -0,0 +1,23 @@
...
+use Illuminate\Support\Facades\Auth;
+
+class PatientInfoService
+{
+    public function searchPatients($searchTerm)
+    {
+        $user = Auth::user();
+        if ($user && $user->sqli_on) {
+            $query = "SELECT * FROM patient WHERE first_name LIKE '%$searchTerm%' " .
+                     "OR last_name LIKE '%$searchTerm%'";
```
alherrera-cs, line 15: Search term is directly inserted, making it vulnerable to SQL injection

app/Http/Controllers/PatientInfoController.php
```diff
@@ -2,28 +2,45 @@
...
+        $searchTerm = $request->input('search', '');
+        $patients = [];
+        $searchPerformed = false;
+        $user = Auth::user();
+
+        try {
+            if (!empty($searchTerm) && $user->sqli_on) {
+                $patients = $this->patientInfoService->searchPatients($searchTerm);
+                $searchPerformed = true;
+            }
```
alherrera-cs, line 30: If a search term is provided and SQL injection is toggled on, execute the vulnerable query to simulate SQL injection.

resources/views/patients/index.blade.php
```diff
@@ -4,34 +4,67 @@
...
+            <div class="p-4 bg-red-100 text-red-800 rounded mb-4">
+                {{ $errors->first('search') }}
+            </div>
+        </div>
+    @endif
+
     <div class="py-12">
-        <div class="max-w-7xl mx-auto sm:px-6 lg:px-8">
-            <div class="bg-white dark:bg-gray-800 overflow-hidden shadow-sm sm:rounded-lg">
+        <!-- Search Form Container (fixed width) -->
```
c-a-ray, line 17: Would you mind looking into making the elements on this page a little more responsive to resizing? I think that will probably mean not using a fixed width for the search form container.

We're just currently in the process of fixing UI resizing issues, so it would be nice to have new stuff coming in not have that problem.

Also, not super important, but the comment "Search Form Container (fixed width)" is a little weird. Maybe just "Patient search form"

resources/views/patients/index.blade.php
```diff
@@ -4,34 +4,67 @@
...
+                                            View Details
+                                        </a>
+                                    </td>
+                                </x-table-row>
+                            @endforeach
+                        </x-table>
+                    </div>
+                @else
+                    <div class="bg-white dark:bg-gray-800 shadow-sm sm:rounded-lg border border-gray-200 dark:border-gray-700 p-6">
+                        <p class="text-gray-800 dark:text-gray-200">No patients found. Please adjust your search criteria.</p>
```
c-a-ray, line 64: Can you center this text?

image

c-a-ray (changes requested):
Nice work. I found a couple of issues, but they should all be pretty quick fixes. Let me know if you want help with anything.

app/Http/Controllers/PatientInfoController.php
```diff
@@ -2,30 +2,53 @@
...
-    public function show($id)
+    public function index(Request $request): View|RedirectResponse
     {
-        $patient = Patient::find($id);
+        $term            = $request->input('search', '');
+        $patients        = [];
+        $searchPerformed = false;
+        $user            = Auth::user();
+
+        if ($term !== '') {
```
c-a-ray, line 39: Bug here. `$term` is `null` when no input is provided, so `$term !== ''` is always true and the search logic always runs (unless you type a literal `''` into the search field). 

Also, when you put nothing in the search input field and click "Search Patients", `null` gets passed as the `$searchTerm` to `searchPatients` (which expects a string), causing the app to crash with this error:


>App\Services\PatientInfoService::searchPatients(): Argument #1 ($searchTerm) must be of type string, null given, called in /var/www/html/app/Http/Controllers/PatientInfoController.php on line 45 

So you just need something to deal with the `null` value in `index`. There are a few ways you could handle this.

Also, consider renaming `$term` to `$searchTerm` or something, just to make it a bit clearer what it is.

resources/views/patients/index.blade.php
```diff
@@ -4,34 +4,83 @@
             {{ __('Patient Information') }}
         </h2>
     </x-slot>
+
```
c-a-ray, line 7: Nice work fixing the UI issue. Looks good now

app/Services/PatientInfoService.php
```diff
@@ -0,0 +1,36 @@
...
+                SELECT *
+                FROM patient
+                WHERE first_name LIKE '%{$searchTerm}%'
+                   OR last_name  LIKE '%{$searchTerm}%'
+            SQL;
+
+            return DB::select($sql);
+        }
+
+        // === SAFE ===
```
c-a-ray, line 30: (Minor) The "===" comments don't really follow the style that we use everywhere else. If you feel that we need these, I think it'd be better to just do `// Vulnerable` and `// Secure`.

But really, I think the best option is to remove them all together. It's already clear from the code and the PHPDoc comment what is happening.

c-a-ray (changes requested):
Nice work fixing the UI issue. Looks much better.

This is really close, but there are at least two issues that we should fix before merging:
1. There's a bug in `PatientInfoController` where `$term` is `null` when you don't put anything in the "Search Patients" input. This causes all of the search logic to run even if no input is entered (I'm pretty sure that's not your intent), and causes the app to crash because `PatientInfoService::searchPatients` expects a string.
2. The vulnerable SQL query is still structured such that the `$searchTerm` is inserted twice, which restricts what you can do with SQL injection quite a bit.

There are a couple other comments, but those two things are the most important. Both are super simple fixes luckily.

c-a-ray (approved):
Looks good. Just need to merge in all of the new changes in master

c-a-ray (approved):
Nice work

PR #133, opened 2025-04-30
A few more demo setup changes

c-a-ray:
Since we're now in the penetration testing phase, we'll probably want to have the demo deployed so we can also test in that environment. When I was going through the GCP deployment process again today, a few things came up that I wanted to change. This PR makes those changes:

- Changes the "prod" environment to "demo". I think this makes more sense since we really don't have a "production" environment. What we've been calling "prod" is really just for the demo.
- Disables the registration feature automatically in the demo environment. When we deployed before, I created a separate branch and deleted all of the stuff having to do with the registration feature. This PR makes it so we don't need to do that, and the registration link/routes are automatically disabled when the environment is 'demo'.
- Filters out Google Cloud health checks from the user activity logs. This is just to keep the logs cleaner. The health checks happen very frequently and clutter things up.
- Makes a small tweak to the user seeding logic. Instead of seeding the dev user only when env is explicitly 'local', I changed it to seed the dev user whenever the env is not 'demo'. The only difference is that the dev user is still created if the environment is set to something like 'test' (I had to do this earlier to test something, so it's not a totally contrived use-case).

### DIFF
--- a/app/Http/Middleware/LogUserActivityMiddleware.php
+++ b/app/Http/Middleware/LogUserActivityMiddleware.php
@@ -20,6 +20,9 @@ class LogUserActivityMiddleware
     public function __invoke(Request $request, Closure $next): Response
     {
         $response = $next($request);
+        if ($request->is('health') && str_contains($request->userAgent(), 'GoogleHC')) {
+            return $response;
+        }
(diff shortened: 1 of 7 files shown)


BrynnMartina (approved):
Nice work!

PR #134, opened 2025-05-05
Small content updates

c-a-ray:
Some small content tweaks:
- Adds "Patient Information" to the SQL injection row in the vuln toggles table. We never did this after adding SQL injection to the patient info page.
- Adds better names for the links on the dashboard page, and removes the "Links" header. Links header just looks janky to me and seems redundant (it's clear they are links without it, and it looks cleaner).


Swangeon (approved):
Looks all good on my end.

BrynnMartina (approved):
Ditto. Looks good

PR #135, opened 2025-05-14
Feedback Search Bug Fix

Swangeon:
The error `The GET method is not supported for route feedback/search. Supported methods: POST.` when you did a search on the Patient Feedback Page and then tried to submit a comment after. This issue came from the fact that, by default, if validation of the data through a validator request class fails, then it tries to send you back to where you were with a GET and the `feedback/search` route isn't setup to be accessed with a GET request.

The main way I solved this was to:

- Override the `failedValidation` method in the [`FeedbackCommentRequest`](https://github.com/BeaverHealth-Vulnerable-Web-App/BeaverHealth-Vulnerable-Web-App/pull/135/files#diff-be2ebdd0ccd20b8d00b1b531de231a0a85bcec9461b2e08e2e6cd9ca9eee6375R35) class to specify a custom redirect route when the validation fails

### DIFF
--- a/app/Http/Requests/FeedbackCommentRequest.php
+++ b/app/Http/Requests/FeedbackCommentRequest.php
@@ -3,6 +3,8 @@
 namespace App\Http\Requests;
 
 use Illuminate\Foundation\Http\FormRequest;
+use Illuminate\Contracts\Validation\Validator;
+use Illuminate\Validation\ValidationException;
 
 class FeedbackCommentRequest extends FormRequest
 {
@@ -25,5 +27,16 @@ class FeedbackCommentRequest extends FormRequest
             'patient_id' => 'required|exists:patient,patient_id',
             'feedback' => 'required|string'
         ];
+    }
+
+    /**
+     * Handle a failed validation attempt.
+     */
+    protected function failedValidation(Validator $validator): void
+    {
+        throw new ValidationException(
+            $validator,
+            redirect()->route('feedback')->withErrors($validator)->withInput()
+        );
     }
 }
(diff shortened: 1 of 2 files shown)


BrynnMartina (dismissed):
Looks good and works on my end!

app/Http/Requests/FeedbackCommentRequest.php
```diff
@@ -26,4 +28,15 @@ public function rules(): array
             'feedback' => 'required|string'
         ];
     }
+
+    /**
+     * Handle a failed validation attempt.
+     */
+    protected function failedValidation(Validator $validator): void
```
c-a-ray, line 35: Nice. This is what's actually solving the problem.

resources/views/components/patient-dropdown.blade.php
```diff
@@ -16,7 +16,9 @@
     <input type="text" id="patient_search" x-model="search" 
            @input="showDropdown = true"
            @focus="showDropdown = true"
-           class="w-full p-2 border border-gray-300 dark:border-gray-600 rounded bg-white dark:bg-gray-700 text-gray-900 dark:text-white" autocomplete="off" />
+           class="w-full p-2 border border-gray-300 dark:border-gray-600 rounded bg-white dark:bg-gray-700 text-gray-900 dark:text-white" 
+           autocomplete="off" 
+           required />
```
c-a-ray, line 21: I get the intent here, but I don't think this is quite right. The form isn't actually submitting the `patient_search` text, but rather the `patient_id` (the hidden input). So all the `required` here does is make it so the button doesn't work if there's no text in the patient search field. You can still enter a character or two (or any invalid input) and the button will then work even with no patient selected. Also don't love how it forces the dropdown to expand.

The thing that is actually solving the problem is the failed validation handler in the form request class. If you didn't have that, you'd still get the error if you put any invalid input into the patient field and clicked the button. The attribute here doesn't address the underlying problem and is kind of misleading (gives the impression that a patient must be selected, but it doesn't actually enforce that).

tl;dr I'd vote to revert this change and just stick with the failed validation handler.

c-a-ray (changes requested):
Nice work figuring this out. Just one thing to look at. No additional work required other than potentially reverting the changes in `patient-dropdown.blade.php`.

c-a-ray (approved):
Good stuff, nice work

PR #137, opened 2025-05-14
UI Tweaks

BrynnMartina:
* Changed the style of the Add Medical Records and Request Medical Records page to look more cohesive with the Patient Feedback page.

Before:
Screenshot 2025-05-13 234537

After:
Screenshot 2025-05-13 233602

* Fixed the HTML from showing sensitive patient info via the dropdown menu.

Before:
Screenshot 2025-05-13 235237

After:
Screenshot 2025-05-13 235353

* Shortened MIME error popup so it doesn't extend the whole page.

Before:
Screenshot 2025-05-14 005752

After:
Screenshot 2025-05-14 005653

* Beautified the login page rate limiter.

Before:
Screenshot 2025-05-13 235658

After:
Screenshot 2025-05-14 003640


app/Http/Controllers/AddRecordsController.php
BrynnMartina, line 1: Shows only ID, name, and SSN in the dev tools html page

app/Http/Middleware/LoginPageRateLimiter.php
BrynnMartina, line 1: Redirects rate limiter notification to 429-error.blade.php

app/Http/Controllers/AddRecordsController.php
```diff
@@ -17,7 +17,7 @@ public function __construct(PatientRecordService $patientRecordService)
 
     public function index()
     {
-        $patients = Patient::all();
+        $patients = Patient::select('patient_id', 'first_name', 'last_name', 'ssn')->get();
```
BrynnMartina, line 20: This line is responsible for limiting patient info in the dev tools html page

app/Http/Middleware/LoginPageRateLimiter.php
```diff
@@ -28,7 +28,11 @@ public function handle(Request $request, Closure $next): Response
                 'Login page access blocked due to too many requests',
                 ['throttle_key' => $key]
             );
-            return response('Too many login page requests. Please try again later.', Response::HTTP_TOO_MANY_REQUESTS);
+            return response()->view(
```
BrynnMartina, line 31: Redirects rate limiter notification to 429-error.blade.php

app/Http/Middleware/LoginPageRateLimiter.php
```diff
@@ -28,7 +28,11 @@ public function handle(Request $request, Closure $next): Response
...
+                'components.429-error',
+                ['retryAfter' => RateLimiter::availableIn($key)],
+                429
+            );
```
BrynnMartina, line 35: Redirects rate limiter notification to 429-error.blade.php

resources/views/components/429-error.blade.php
```diff
@@ -0,0 +1,11 @@
+<x-guest-layout>
```
c-a-ray, line 1: Already talked about this in a comment above, but I think if we're going to render anything for the lockout, we should either use a cached static file or inline HTML (or something like that).

c-a-ray (changes requested):
Nice work, just a couple things:
- Two little code style things (not really important)
- The new MIME type error doesn't really make sense
- A discussion about whether/how we want to render content for the login page lockout

c-a-ray (approved):
Nice work. One suggestion, but approving anyway because it's not a big deal

c-a-ray (approved):
Nice!

PR #138, opened 2025-05-21
Make rate limiters configurable

c-a-ray:
In our last meeting we decided to disable rate limiting for login page accesses locally. I took this a little farther and made it so:
- Both the login page access rate limiter and the login attempt rate limiter are configurable in the `local` environment
  - By default, both are disabled locally
  - They can be enabled by setting environment variables `LOGIN_PAGE_ACCESS_RATE_ENABLE` and `LOGIN_ATTEMPTS_RATE_ENABLE`, or by editing the default values in `config/auth.php`
- Both rate limiters are still always enabled in the `demo` environment
- Also made max attempts and hit decay rate configurable for both rate limiters. These can be configured regardless of environment

While I was at it, I also:
- Added inline HTML and styling to the lockout response, so the lockout view now looks nicer and shows the time remaining, but uses less resources than rendering a Blade template.
- Added tests to make sure the configuration logic works as expected
- Added a hash of the client's user agent to the key for the login page access rate limiter. In current `master`, it only uses the IP, which I think could potentially be a problem for people with the same IP. For example, if multiple people were on the same VPN or local network and had the same IP, and one person triggers the lockout, it would lock everybody with that IP out. Including the user agent in the key is just a way to prevent that. It's not foolproof, but I haven't been able to find a perfect solution. I think a better solution would probably be to add a captcha, but that would require more development effort, and I think this works well enough, especially since it only matters for the demo.

### DIFF
--- a/app/Http/Controllers/AuthenticatedSessionController.php
+++ b/app/Http/Controllers/AuthenticatedSessionController.php
@@ -11,6 +11,18 @@ use Illuminate\View\View;
 
 class AuthenticatedSessionController extends Controller
 {
+    private readonly bool $rateLimitEnabled;
+
+    /**
+     * Create a new controller instance and determine whether requests should be rate-limited.
+     */
+    public function __construct()
+    {
+        $env = config('app.env');
+        $enableLocalRateLimit = config('auth.login_attempts_rate_limit.enable_locally');
+        $this->rateLimitEnabled = $env === 'demo' || ($env === 'local' && $enableLocalRateLimit);
+    }
+
     /**
      * Display the login view.
      */
@@ -24,7 +36,12 @@ class AuthenticatedSessionController extends Controller
      */
     public function login(LoginRequest $request): RedirectResponse
     {
-        $request->authenticate();
+        if ($this->rateLimitEnabled) {
+            $request->authenticateOrThrottle();
+        } else {
+            $request->authenticate();
+        }
+
         $request->session()->regenerate();
         return redirect()->intended(route('dashboard', absolute: false));
     }
--- a/app/Http/Middleware/LoginPageRateLimiter.php
+++ b/app/Http/Middleware/LoginPageRateLimiter.php
@@ -111,7 +111,7 @@ class LoginPageRateLimiter
                         border-radius: 0.5rem;
                         box-shadow: 0 2px 10px rgba(0,0,0,0.2);
                         text-align: center;
-                        max-width: 24rem;
+                        max-width: 20rem;
                     }
                     .card h2 {
                         color: #f56565;
--- a/app/Http/Middleware/LoginPageRateLimiter.php
+++ b/app/Http/Middleware/LoginPageRateLimiter.php
@@ -10,6 +10,28 @@ use Symfony\Component\HttpFoundation\Response;
 
 class LoginPageRateLimiter
 {
+    // Length of the User-Agent hash prefix used in the rate limit key
+    private const USER_AGENT_HASH_LENGTH = 12;
+
+    private readonly bool $enabled;
+    private readonly int $maxAttempts;
+    private readonly int $hitDecay;
+
+    /**
+     * Creates a new LoginPageRateLimiter instance.
+     *
+     * Initializes rate limiting thresholds from configuration.
+     */
+    public function __construct()
+    {
+        $env = config('app.env');
+        $enableLocally = config('auth.login_page_access_rate_limit.enable_locally');
+
+        $this->enabled = $env === 'demo' || ($env === 'local' && $enableLocally);
+        $this->maxAttempts = config('auth.login_page_access_rate_limit.max_attempts');
+        $this->hitDecay = config('auth.login_page_access_rate_limit.decay_seconds');
+    }
+
     /**
      * Throttles repeated access to the login page based on IP address.
      *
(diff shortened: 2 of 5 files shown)


app/Http/Controllers/AuthenticatedSessionController.php
```diff
@@ -11,6 +11,18 @@
 
 class AuthenticatedSessionController extends Controller
 {
+    private readonly bool $rateLimitEnabled;
```
c-a-ray, line 14: I just learned about this `readonly` thing. It makes it so you can only set the value once (which is exactly what we want here).

app/Http/Middleware/LoginPageRateLimiter.php
```diff
@@ -10,6 +10,28 @@
 
 class LoginPageRateLimiter
 {
+    // Length of the User-Agent hash prefix used in the rate limit key
+    private const USER_AGENT_HASH_LENGTH = 12;
```
c-a-ray, line 14: This is a constant rather than a configurable variable because it's more of an implementation detail than something you'd want to configure.

app/Http/Requests/Auth/LoginRequest.php
```diff
@@ -69,9 +90,9 @@ public function authenticate(): void
      *
      * @throws ValidationException If too many attempts have been made
      */
-    protected function ensureIsNotRateLimited(): void
+    private function ensureIsNotRateLimited(): void
```
c-a-ray, line 93: Made this private because there are no children of this class, so there's no reason for it to be `protected`

app/Http/Requests/Auth/LoginRequest.php
```diff
@@ -80,19 +101,17 @@ protected function ensureIsNotRateLimited(): void
             ['throttle_key' => $this->throttleKey()]
         );
 
-        event(new Lockout($this));
```
c-a-ray, line 83: This wasn't actually doing anything. We don't have anything listening for this event, so it's pointless.

app/Http/Middleware/LoginPageRateLimiter.php
```diff
@@ -21,18 +43,92 @@ class LoginPageRateLimiter
...
+    /**
+     * Generate the rate-limiting key for this login attempt.
+     *
+     * @param Request $request
+     * @return string The key used for throttling (client IP + hash of User Agent)
+     */
+    private function throttleKey(Request $request): string
+    {
+        $ip = $request->ip();
+        $userAgentHash = substr(md5($request->userAgent()), 0, self::USER_AGENT_HASH_LENGTH);
```
c-a-ray, line 75: See PR description for why this now includes a hash of the user agent

tests/Feature/AuthFeatureTest.php
```diff
@@ -277,29 +264,120 @@ public function testAuthenticatedSessionPersistsAcrossRequests(): void
         $this->assertAuthenticatedAs($user);
     }
 
-    public function testLoginLockoutAfterFailedAttempts(): void
+    public function testDemoEnvRateLimitsLoginAttempts(): void
```
c-a-ray, line 267: These new tests just test that the new configuration options work as expected

app/Http/Requests/Auth/LoginRequest.php
alherrera-cs: Good catch!

alherrera-cs (dismissed):
Looks good to me. The config flexibility seems really useful, and the user-agent hash addition for the demo makes sense. Thanks for making these changes!

config/auth.php
```diff
@@ -23,4 +23,28 @@
         ],
     ],
 
+    // Rate limiting for login page accesses
+    'login_page_access_rate_limit' => [
+        // Whether to enable rate limiting for login page accesses locally
+        'enable_locally' => env('LOGIN_PAGE_ACCESS_RATE_ENABLE', false),
+
+        // The number of accesses before a lockout occurs
+        'max_attempts' => env('LOGIN_PAGE_ACCESS_RATE_MAX', 15),
```
BrynnMartina, line 32: Just curious as to your motivation behind changing it to 15

BrynnMartina, line 32: Just curious as to why you changed it to 15

app/Http/Middleware/LoginPageRateLimiter.php
```diff
@@ -21,18 +43,92 @@ class LoginPageRateLimiter
...
+                        color: #f56565;
+                        margin-bottom: 1rem;
+                    }
+                </style>
+            </head>
+            <body>
+                <div class="card">
+                    <h2>Too Many Requests</h2>
+                    <p>You’ve hit the login page too many times in a short period.</p>
+                    <p>Please try again in {$waitSeconds} {$pluralized}.</p>
```
BrynnMartina, line 126: Nice touch

app/Http/Middleware/LoginPageRateLimiter.php
```diff
@@ -21,18 +43,92 @@ class LoginPageRateLimiter
...
+                        height: 100vh;
+                        margin: 0;
+                    }
+                    .card {
+                        background-color: #2d3748;
+                        padding: 2rem;
+                        border-radius: 0.5rem;
+                        box-shadow: 0 2px 10px rgba(0,0,0,0.2);
+                        text-align: center;
+                        max-width: 24rem;
```
BrynnMartina, line 114: Very very minimal - the current message has this on the first line "You've hit the login page too many times in a short" and "period." alone on a second line. I changed max-width to 20rem instead of 24rem so there are no lonely words, so it looks like this:
image

BrynnMartina (commented):
I think this looks Cody. Just one very minor UI tweak if you felt like it.

config/auth.php
c-a-ray: Just to try to make it less likely that someone hits the rate limit accidentally. I could change it back to 10 if you think that's better, doesn't matter too much to me

app/Http/Middleware/LoginPageRateLimiter.php
c-a-ray: Updated to `20rem`

BrynnMartina (approved):
Looks great!

PR #145, opened 2025-05-22
Patient Information Page: Feature Tests

alherrera-cs:
Description:
This PR adds a new feature test suite for the Patient Info page. It covers key scenarios including:

-     Page load and basic access control
-     Valid and invalid search behavior
-     SQLi toggle behavior
-     “View Details” link rendering
-     Edge cases like empty or unmatched search queries

Please let me know if anything looks off or if you'd like more coverage somewhere.

### DIFF
--- a/tests/Feature/PatientInfoFeatureTest.php
+++ b/tests/Feature/PatientInfoFeatureTest.php
@@ -1,0 +1,126 @@
+<?php
+
+namespace Tests\Feature;
+
+use App\Models\User;
+use Illuminate\Support\Facades\Route;
+use Tests\TestCase;
+
+class PatientInfoFeatureTest extends TestCase
+{
+    private User $user;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->user = User::factory()->create(['view_patient_info' => true]);
+        $this->actingAs($this->user);
+    }
+
+    public function testPageLoads(): void
+    {
+        $response = $this->get(route('patients.index'));
+        $response->assertOk();
+        $response->assertSee('Patient Information');
+    }
+
+    public function testSearchReturnsPatients(): void
+    {
+        $user = User::factory()->create(['view_patient_info' => true]);
+        $this->actingAs($user);
+
+        $patient = \App\Models\Patient::create([
+            'first_name' => 'John',
+            'last_name' => 'Smith',
+            'date_of_birth' => '1985-04-01',
+            'policy_number' => 'ABC123',
+            'address' => '123 Main St',
+            'is_employee' => false,
+            'ssn' => '123-45-6789',
+        ]);
+
+        $this->get(route('patients.index', ['search' => 'John']))
+            ->assertOk()
+            ->assertSeeText('John Smith')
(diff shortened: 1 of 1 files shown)


Swangeon (changes requested):
What you have looks good, there are several things I would add to it though. 

1. One thing that I would add is a test for proper error handling of a different type of input like a raw integer. 
2. Test for SQL injection when the `sqli_on` toggle is disabled.
3. Test how you want to handle super long inputs (like 1000 character input which can be done with `str_repeat`)

tests/Feature/PatientInfoFeatureTest.php
```diff
@@ -0,0 +1,151 @@
...
+        $response->assertOk();
+        $response->assertSee('Patient Information');
+    }
+
+    public function testSearchReturnsPatients(): void
+    {
+        $user = User::factory()->create(['view_patient_info' => true]);
+        $this->actingAs($user);
+
+        $patient = \App\Models\Patient::create([
```
BrynnMartina, line 32: Instead of referenceing\App\Models here, you can put it up top so it's `use App\Models\Patient` and then on line 32 just do `$patient = Patient::create`

Additionally, you're using `Patient::create(...)` here when I think `Patient::factory()->create(...)` might work better. You're using it in other places already.

tests/Feature/PatientInfoFeatureTest.php
```diff
@@ -0,0 +1,151 @@
...
+    public function testUnauthenticatedRedirect(): void
+    {
+        auth()->logout();
+        $this->get(route('patients.index'))->assertRedirect(route('login'));
+    }
+
+    public function testAccessDeniedWithoutPermission(): void
+    {
+        $user = User::factory()->create(['view_patient_info' => false]);
+        $this->actingAs($user);
```
BrynnMartina, line 79: This is optional, I see you run these two lines a lot in your functions so you could create a helper function instead, like
`private function actingAsUserWithPermission(bool $hasPermission = true): void
{
    $user = User::factory()->create(['view_patient_info' => $hasPermission]);
    $this->actingAs($user);
}`

tests/Feature/PatientInfoFeatureTest.php
```diff
@@ -0,0 +1,151 @@
...
+        $this->get(route('patients.index', ['search' => 'NoSuchName']))
+            ->assertOk()
+            ->assertSeeText('No patients found');
+    }
+    public function testPatientSearchShowsDetailsLink(): void
+    {
+        $user = User::factory()->create(['view_patient_info' => true]);
+        $this->actingAs($user);
+
+        $patient = \App\Models\Patient::create([
```
BrynnMartina, line 111: Another place where you could move `\App\Models`

BrynnMartina (commented):
I agree with Sean's tips with expanding the testing a bit more. I picked out a few minimal coding inconsistencies, but it all passes on my end!

c-a-ray (commented):
Looks like Sean and Brynn already have this covered. Don't think I have anything else to add

PR #146, opened 2025-05-22
File Upload Feature Tests

BrynnMartina:


alherrera-cs (commented):
Nice start! I think this covers the main stuff really well. 
Couple quick ideas if you’re planning to build it out more:

- Maybe test what happens if file_upload_on is false? Just to make sure uploads are blocked as expected.

- Could be good to check how it handles duplicate filenames.

- Maybe test how it handles files that look normal but aren’t, like a .pdf that isn’t actually a real PDF (mime type mismatch). Just to make sure it’s not letting weird or fake files through.

Just a couple suggestions. Looks great so far!

tests/Feature/AddRecordsFeatureTest.php
```diff
@@ -0,0 +1,117 @@
...
+
+namespace Tests\Feature;
+
+use Illuminate\Http\UploadedFile;
+use Illuminate\Support\Facades\Storage;
+use Tests\TestCase;
+use App\Models\User;
+use App\Models\Patient;
+
+class AddRecordsUploadTest extends TestCase
```
c-a-ray, line 11: Class name seems a little redundant. No biggie though, no reason to hold this up

c-a-ray (dismissed):
Nice job. Agree with Alexa that there are a handful of other things it would be good to test, but this at least gets us some test coverage. Since we're right at the edge of code freeze I don't see any reason to hold this up.

tests/Feature/AddRecordsFeatureTest.php
```diff
@@ -0,0 +1,117 @@
...
+
+namespace Tests\Feature;
+
+use Illuminate\Http\UploadedFile;
+use Illuminate\Support\Facades\Storage;
+use Tests\TestCase;
+use App\Models\User;
+use App\Models\Patient;
+
+class AddRecordsFeatureTest extends TestCase
```
BrynnMartina, line 11: Changed the class name to match the file name, as the other feature tests do the same

c-a-ray (approved):
Looks good

PR #148, opened 2025-05-28
Fix logout force browse bug

c-a-ray:
This PR fixes the bug that Sean found where you get a `MethodNotAllowedHttpException` when navigating directly to `/logout`.

It should actually fix any case where the wrong HTTP method is used for any registered route. It does this by adding an app-wide exception handler for `MethodNotAllowedHttpException`.

We already have handling for `NotFoundHttpException`, so we should now be covered for all cases where a user either tries to access an unregistered route, or tries to use the wrong HTTP method for a registered route.

### DIFF
--- a/bootstrap/app.php
+++ b/bootstrap/app.php
@@ -1,6 +1,7 @@
 <?php
 
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpKernel\Exception\MethodNotAllowedHttpException;
 use Illuminate\Foundation\Application;
 use Illuminate\Foundation\Configuration\Exceptions;
 use Illuminate\Foundation\Configuration\Middleware;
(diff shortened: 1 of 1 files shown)


Swangeon (commented):
Was there no way to send the user back to where they were? The `Referer` header is there so could we use that?

c-a-ray:
> Was there no way to send the user back to where they were? The `Referer` header is there so could we use that?

Not really. Or at least not easily. The `Referer` header isn't available when users type URLs directly into the address bar (which is the main way this exception would happen). It's only present when navigating via links or forms, from what I understand.

I also found out that session information isn't available at this point in the request lifecycle (because the session middleware hasn't run yet when these exceptions are thrown), so we can't access the previous URL that Laravel stores in the session. 

That made me realize that `Auth::check()` always returns false in these exception handlers, so I simplified both of them to just redirect to `/login`. If the user is authenticated, that will automatically redirect to `/dashboard`.

Swangeon (approved):
Makes sense, was hoping there was something we could do with the middleware or `Referer`. I'm good to go on this then.

PR #149, opened 2025-05-28
Add pagination to Patient Feedback page

c-a-ray:
This PR fixes basically all issues on the Patient Feedback page found during testing.

The main thing it does is add pagination to the comments list. Instead of loading all comments at once, it now loads 10 at a time. This fixes the long load times and server errors that occur when there is a large amount of comments data.

While I was doing this, I also fixed a few other issues:
- There was no wrapping when strings were really long. Updated styling to make it so long strings are wrapped.
- Changed the `/feedback/search` route to a GET method. It currently uses the POST method, which is semantically incorrect for search operations. It really should be a GET since it's reading/filtering data rather than modifying server state. This change was needed to enable proper pagination functionality, as pagination links use GET requests and were incompatible with the POST search form.
- Adds some frontend validation to prevent super long strings in the Search Comments field. Did this to fix a "Request-URI Too Long" error.

### DIFF
--- a/resources/views/feedback/index.blade.php
+++ b/resources/views/feedback/index.blade.php
@@ -39,8 +39,13 @@
                         <form action="{{ route('feedback.search') }}" method="GET">
                             <div class="mb-4">
                                 <label for="search_name" class="block mb-2 font-semibold">Search by Name</label>
-                                <input type="text" class="w-full p-2 border border-gray-300 rounded text-base mb-4 dark:text-white dark:bg-gray-900 dark:border-gray-600"
-                                    id="search_name" name="search_name" value="{{ $search_name ?? '' }}" required>
+                                <input type="text"
+                                    class="w-full p-2 border border-gray-300 rounded text-base mb-4 dark:text-white dark:bg-gray-900 dark:border-gray-600"
+                                    id="search_name"
+                                    name="search_name"
+                                    value="{{ $search_name ?? '' }}"
+                                    maxlength="255"
+                                    required>
                             </div>
                             <button type="submit" class="w-full md:w-auto px-6 py-4 bg-gray-500 text-white font-medium text-sm leading-tight uppercase rounded shadow-md hover:bg-gray-600 hover:shadow-lg focus:bg-gray-600 focus:shadow-lg focus:outline-none focus:ring-0 active:bg-gray-700 active:shadow-lg transition duration-150 ease-in-out dark:bg-gray-700 dark:hover:bg-gray-600">
                                 Search
--- a/resources/views/feedback/index.blade.php
+++ b/resources/views/feedback/index.blade.php
@@ -83,7 +83,7 @@
                                 {{ $comment->patient->first_name }} {{ $comment->patient->last_name }}
                             @endif
                         </h5>
-                        <p class="text-gray-700 dark:text-gray-300" style="word-break: break-all; overflow-wrap: anywhere; max-width: 100%; hyphens: auto;">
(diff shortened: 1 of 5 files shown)


resources/views/feedback/index.blade.php
```diff
@@ -36,11 +36,16 @@
                         Search Comments
                     </div>
                     <div class="p-4">
-                        <form action="{{ route('feedback.search') }}" method="POST">
-                            @csrf
+                        <form action="{{ route('feedback.search') }}" method="GET">
                             <div class="mb-4">
                                 <label for="search_name" class="block mb-2 font-semibold">Search by Name</label>
-                                <input type="text" class="w-full p-2 border border-gray-300 rounded text-base mb-4 dark:text-white dark:bg-gray-900 dark:border-gray-600" id="search_name" name="search_name" required>
+                                <input type="text"
```
c-a-ray, line 42: The only thing that changed here is the `maxlength="255"`. I just broke it onto multiple lines to make it easier to read

resources/views/feedback/index.blade.php
```diff
@@ -69,24 +76,36 @@
...
-                    <small class="text-gray-600 dark:text-gray-400">Posted on: {{ $comment->created_at->format('M d, Y H:i') }}</small>
-                </div>
+                    @foreach($feedback as $comment)
+                    <div class="border-b border-gray-200 dark:border-gray-700 mb-4 pb-4">
+                        <h5 class="mb-2 text-lg font-medium dark:text-white dark:bg-gray-800">
+                            @if($comment->patient)
+                                {{ $comment->patient->first_name }} {{ $comment->patient->last_name }}
+                            @endif
+                        </h5>
+                        <p class="text-gray-700 dark:text-gray-300" style="word-break: break-all; overflow-wrap: anywhere; max-width: 100%; hyphens: auto;">
```
c-a-ray, line 86: I couldn't get this to work with just Tailwind, so I used CSS. If anybody knows how to get it to work with Tailwind, would be happy to change it, but I'm not sure it's possible

routes/web.php
```diff
@@ -51,7 +51,7 @@ function () {
 
         Route::get('/feedback', [FeedbackController::class, 'index'])->name('feedback');
         Route::post('/feedback/store', [FeedbackController::class, 'store'])->name('feedback.store');
-        Route::post('/feedback/search', [FeedbackController::class, 'search'])->name('feedback.search');
+        Route::get('/feedback/search', [FeedbackController::class, 'search'])->name('feedback.search');
```
c-a-ray, line 54: POST should really only be used for creating/modifying data. This is strictly a read operation, so it should be a GET

tests/Feature/FeedbackFeatureTest.php
```diff
@@ -33,12 +33,7 @@ private function sendTestStoreFeedback($comment, $patientID = 1000): TestRespons
 
     private function sendTestSearchFeedback($name): TestResponse
     {
-        return $this->postWithCsrf(
-            route('feedback.search'),
-            [
-                'search_name' => $name
-            ]
-        );
+        return $this->get(route('feedback.search', ['search_name' => $name]));
```
c-a-ray, line 36: Changes were needed because of the switch from POST to GET

Swangeon (changes requested):
I was able to find out the tailwind problem with the word wrapping but nonetheless looks good

resources/views/feedback/index.blade.php
c-a-ray: I had to make some changes to this but it got me on the right track, thanks. Now uses tailwind

Swangeon (approved):
Looks good to go for me!

PR #151, opened 2025-05-29
Command Injection Fix

BrynnMartina:
Fixes command injection so when the file upload page is vulnerable, command injection still works even if a patient doesn't have any files associated with their name. 

Also fixed the dark background cutting out (white background showing).


c-a-ray (approved):
Looks good

